PR #8, opened 2022-06-27
BKNDLSS-28515 Snackbar

addKiwi:
image
image
image
image
image
image


v-excelsior (changes requested):

components/bl-snackbar-component/src/index.js
```diff
@@ -0,0 +1,98 @@
...
+      component.show = false
+    }
+  }, [component.show])
+
+  useEffect(() => {
+    if (visible) {
+      addClasses('bl-customComponent-snackbar_show')
+    } else {
+      removeClasses('bl-customComponent-snackbar_show')
+    }
```
v-excelsior, line 51: We could pass visible into useClasses, and it should be redundant.

components/bl-snackbar-component/src/helpers/useClasses.js
```diff
@@ -0,0 +1,33 @@
+import { useMemo } from 'react'
+
+export const useClasses = (horizontalPosition, verticalPosition, type, visible) => {
+  return useMemo(() => {
+    let text = 'bl-customComponent-snackbar'
```
v-excelsior, line 5:
```
const classes = ['bl-customComponent-snackbar']

if (foo > bar) {
      classes.push('some class')
}

return classes.join(' ')
```

components/bl-snackbar-component/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+    onClose,
+    onAction,
+  } = eventHandlers
+
+  const classes = useClasses(horizontalPosition, verticalPosition, type, visible)
+
+  const Icon = IconsMap[type]
+
+  useEffect(() => {
+    if (autoHide) {
```
v-excelsior, line 39: `if (autoHive && visible)`

components/bl-snackbar-component/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+      if (visible) {
+        setTimeout(() => {
```
v-excelsior, line 41: if one row?

components/bl-snackbar-component/src/helpers/useClasses.js
```diff
@@ -0,0 +1,33 @@
+import { useMemo } from 'react'
```
v-excelsior, line 1: rename file use-classes.js

components/bl-snackbar-component/src/components/button.js
```diff
@@ -0,0 +1,9 @@
+export const Action = ({ className, onClick, content }) => {
+  return (
+    <button
+      className={ className }
+      onClick={ onClick }>
+      { content }
```
v-excelsior, line 6: { children }

components/bl-snackbar-component/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+        { Icon && (
+          <Icon/>
+        ) }
+        <div className="content">
+          { snackContent }
+        </div>
+      </div>
+      <div className="buttons">
+        { showAction && (
+          <Action
```
v-excelsior, line 67:
```
<Action
  className="action"
  onClick={ onAction }>
  { actionContent }
</Action>
```

components/bl-snackbar-component/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+            className="action"
+            onClick={ onAction }
+            content={ actionContent }
+          />
+        ) }
+        { showClose && (
+          <Action
```
v-excelsior, line 74:
```
<Action
  className="close"
  onClick={ close }>
  <Close>
</Action>
```

components/bl-snackbar-component/src/components/icons/close.js
```diff
@@ -0,0 +1,13 @@
+export const Close = () => {
```
v-excelsior, line 1: file name: close, as it under icons directory
component name: CloseIcon - because we import it somewhere.

v-excelsior (changes requested):

components/bl-snackbar-component/README.md
```diff
@@ -0,0 +1,114 @@
...
+| Vertical Position       | Top             | This option is to set the vertical position where snackbar will appear. Available 3 positions: `Top`, `Bottom` and `Center`.                                 |
+| Horizontal Position     | Right           | This option is to set the horizontal position where snackbar will appear. Available 3 positions: `Left`, `Center` and `Right`.                               |
+| Snackbar Content        | "Snackbar info" | This option is to set the snackbar text.                                                                                                                     |
+| Action Button Content   | "Action"        | This option is to set the action button text. Works only if "Show Action Button" is `true`.                                                                  |
+| Type                    | Default         | This option is to set the type of snackbar. Available 5 types: `Info`, `Success`, `Warning`, `Error` and `Default`.                                          |
```
v-excelsior, line 72: At description  should be values, which we could return from handler.

components/bl-snackbar-component/component.json
```diff
@@ -0,0 +1,151 @@
...
+  ],
+  "eventHandlers": [
+    {
+      "name": "onClose",
+      "label": "onClose Click",
+      "handlerDescription": "This event is fired when the user clicks on the close button."
+    },
+    {
+      "name": "onAction",
+      "label": "onAction Click",
```
Valodya, line 134: тоже что и с кликом

components/bl-snackbar-component/src/components/button.js
```diff
@@ -0,0 +1,9 @@
+export const Action = ({ className, onClick, children }) => {
+  return (
+    <button
```
Valodya, line 3: into single line

components/bl-snackbar-component/src/components/button.js
```diff
@@ -0,0 +1,9 @@
+export const Action = ({ className, onClick, children }) => {
```
Valodya, line 1: function declaration here and everywhere

components/bl-snackbar-component/src/helpers/type.js
```diff
@@ -0,0 +1,8 @@
+import { Error, Info, Success, Warning } from '../components/icons'
+
+export const IconsMap = {
```
Valodya, line 3: move it into the `/src/icons/index.js` file

Valodya:
@v-excelsior @addKiwi could you please describe how is it supposed to work in UIBuilder? 

Let's consider a specific case, on my page I have several places where I would like to show alerts, 
- "Sorry, but you can not create a new order because {error from the server}"
- "Your order was created, we contact you asap..."

do I need to place two instances of the UI component on my page?

addKiwi:
@Valodya we can add data binding to `Snackbar Content Logic` and `Type Logic`. Then when we call `Visibility Snackbar`, also setting text and type.
image
image
Or i will modify `Snackbar Content Logic`, which will have optional parameters for component.

Valodya:
From my point of view, the component should be only one on the page, and then we can show any alert and as many different alerts as we need

for that:
1) such handlers "Type/Content" doesn't make sense because it forces us to have only one alert (snackBar) at the same time
2) the component should render a list of alerts, where each one has its own type/text/icon/etc
3) there should be an action to display a new alert from any component logic on the page

does it make sense?

UI Builder - UIComponentsTest - Backendless 2022-07-23 09-36-56

UI Builder - UIComponentsTest - Backendless 2022-07-23 09-36-40

Monosnap 2022-07-21 15-23-05

components/bl-snackbar-component/src/use-classes.js
```diff
@@ -0,0 +1,25 @@
+import { useMemo } from 'react';
+
+export const useClasses = (horizontalPosition, verticalPosition) => {
```
serhmel, line 3: function declaration?

components/bl-snackbar-component/src/use-classes.js
```diff
@@ -0,0 +1,25 @@
+import { useMemo } from 'react';
```
serhmel, line 1: could hooks be placed in helpers folder?

components/bl-snackbar-component/src/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+  if (!display) {
+    return null;
+  }
+
+  return (
+    <div className={ cn('bl-customComponent-snackbar',
+      {
+        'bl-customComponent-snackbar_left'   : horizontalPosition === 'left',
+        'bl-customComponent-snackbar_centerX': horizontalPosition === 'center',
```
v-excelsior, line 82: What if we choose center+center? Does it work?

v-excelsior (changes requested):

components/bl-snackbar-component/src/index.js
addKiwi: Yes, it will work, but not correct. I will fix it (problem with classes).

PR #25, opened 2022-08-03
MARKET-1305  Lottie Animation

StasivPavlo:


addKiwi (commented):
Don't create an external directory for 1-2 file
https://github.com/Backendless/ui-builder-library/wiki/Files#general-rules
I think we don't need a `lib` folder

components/bl-lottieAnimation-component/README.md
```diff
@@ -0,0 +1,43 @@
...
+<dd>Allows to specify looping or non-looping animation. Default value 'true'</dd>
+<dt>Stopped</dt>
+<dd>Allows to specify stopped or not stopped animation.</dd>
+</dl>
+
+### Event Handlers and Bindable Properties
+
+<dl>
+<dt>Data Logic</dt>
+<dd>Allows to specify data for animation in Backendless logic</dd>
```
v-excelsior, line 25: Could we add a little instruction with links where user could find/create an animation?

components/bl-lottieAnimation-component/src/index.js
```diff
@@ -0,0 +1,60 @@
...
+    if (isStopped) {
+      stop();
+    }
+  }, [data]);
+
+  component.playAnimation = () => {
+    play();
+  };
+
+  component.stopAnimation = () => {
```
v-excelsior, line 24: What if we add an action to change the current animation? I know we could do it through data binding but action also could be useful.

components/bl-lottieAnimation-component/src/index.js
```diff
@@ -0,0 +1,60 @@
+import { useEffect, useRef } from 'react';
+import { destroy, loadAnimation, play, stop } from './lottie.min';
+
+const { cn } = BackendlessUI.CSSUtils;
+
+export default function lottieAnimation({ component, eventHandlers }) {
+  const { display, classList, data, isLoop, type, isStopped } = component;
+  const { onHover, onUnhover } = eventHandlers;
```
v-excelsior, line 8: Add onClick event please

v-excelsior (changes requested):

components/bl-lottieAnimation-component/src/index.js
```diff
@@ -0,0 +1,66 @@
+import { useEffect, useRef } from 'react';
+import { destroy, loadAnimation, play, stop } from './lottie.min';
+
+const { cn } = BackendlessUI.CSSUtils;
+
+export default function lottieAnimation({ component, eventHandlers }) {
```
Valodya, line 6: `LottieAnimation`

components/bl-lottieAnimation-component/README.md
Valodya: @StasivPavlo ?

PR #28, opened 2022-08-10
MARKET-1307 Social Login Button 

Saharij:
Screenshot 2022-08-10 at 17 33 55
Screenshot 2022-08-10 at 18 15 27
Screenshot 2022-08-10 at 18 14 12


v-excelsior (changes requested):

components/bl-social-login-button-component/component.json
```diff
@@ -0,0 +1,48 @@
...
+  "properties": [
+    {
+      "type": "select",
+      "name": "buttonOption",
+      "label": "Button option",
+      "showInSettings": true,
+      "defaultValue": "googleplus",
+      "hasLogicHandler": false,
+      "handlerId": "buttonOptionLogic",
+      "options": [
```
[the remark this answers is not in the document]
Saharij: I thought this would be a good solution as the user can choose both the order of the buttons and the number of them.
If the ability to choose the number of buttons should be mandatory, then i can try to do it by checkboxes, what do you think?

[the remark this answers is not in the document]
v-excelsior: Yes, checkboxes will be good. In a case when user want's to change order, he may duplicate component. But the basic behavior must be - one component = multiple logins.

components/bl-social-login-button-component/src/icons-map.js
```diff
@@ -0,0 +1,78 @@
+export const IconsMap = ({ socialNet }) => {
```
v-excelsior, line 1: I mean not rename but make it an object :)

components/bl-social-login-button-component/src/index.js
```diff
@@ -0,0 +1,22 @@
...
+  const { display, classList, disable, showButtonIcon, googleplus, facebook, twitter, linkedin, github } = component;
+  const { onClick } = eventHandlers;
+
+  if (!display) {
+    return null;
+  }
+
+  return (
+    <div className={ cn('bl-customComponent-socialLoginButton', classList, { disable }) }>
+      { googleplus && <Button socialNet={ "googleplus" } showButtonIcon={ showButtonIcon } onClick={ onClick } /> }
```
v-excelsior, line 15: We must pass which type of login used in onClick, also that mean need to add context block 'Used Login' or 'Login Type' at this handler.

components/bl-social-login-button-component/src/index.js
```diff
@@ -0,0 +1,22 @@
...
+      { facebook && <Button socialNet={ "facebook" } showButtonIcon={ showButtonIcon } onClick={ onClick } /> }
```
v-excelsior, line 16: socialNet => providerName or providerType or provider

components/bl-social-login-button-component/src/button.js
```diff
@@ -0,0 +1,46 @@
...
+        const redirectAfterLoginUrl = redirectToPage
+          ? window.location.href.split('?')[0] + `?page=${ redirectToPage }`
+          : window.location.href
+
+        window.location = await Backendless.UserService
+          .getAuthorizationUrlLink(providerCode, fieldsMappings, scope, false, redirectAfterLoginUrl, callbackUrlDomain)
+      }
+    }
+
+    await socialLogin(socialNet, null, null, null, null, null)
```
v-excelsior, line 27: Please ask @Valodya if we must add some properties which could be accepted by Social Login at SDK. I a'm about `fieldsMappings, scope, options, redirectToPage, callbackUrlDomain`

Valodya: I assume it would be helpful

Saharij: I added these options and got an error (fieldsMappings is not defined)

components/bl-social-login-button-component/src/button.js
```diff
@@ -0,0 +1,49 @@
...
+
+        if (redirectToPage) {
+          BackendlessUI.Navigator.goToPage(redirectToPage)
+        }
+      } else {
+        const redirectAfterLoginUrl = redirectToPage
+          ? window.location.href.split('?')[0] + `?page=${ redirectToPage }`
+          : window.location.href
+
+        window.location = await Backendless.UserService.getAuthorizationUrlLink(
```
v-excelsior, line 22: What if we get an error returned from request?

components/bl-social-login-button-component/src/index.js
```diff
@@ -0,0 +1,22 @@
...
+export default function SocialLoginButtonComponent({ component, eventHandlers }) {
+  const { display, classList, disable, showButtonIcon, googleplus, facebook, twitter, linkedin, github } = component;
+  const { onLogin } = eventHandlers;
+
+  if (!display) {
+    return null;
+  }
+
+  return (
+    <div className={ cn('bl-customComponent-socialLoginButton', classList, { disable }) }>
```
v-excelsior, line 14: const `providersList = ['google',...]`

and here:

```
providersList.map(provider => ({ component[provider] && <Button provider={provider}.../>})
```

What about it?

components/bl-social-login-button-component/src/button.js
```diff
@@ -0,0 +1,49 @@
+import { iconsMap } from './icons-map';
+
+const { cn } = BackendlessUI.CSSUtils;
+
+export function Button({ provider, showButtonIcon, onLogin }) {
+  const handleClick = async () => {
```
v-excelsior, line 6: Could we add some error handling?

components/bl-social-login-button-component/src/button.js
```diff
@@ -0,0 +1,49 @@
...
+          .socialLogin(providerCode, fieldsMappings, scope, options, callbackUrlDomain)
+
+        Backendless.UserService.setCurrentUser(user, true)
+
+        if (redirectToPage) {
+          BackendlessUI.Navigator.goToPage(redirectToPage)
+        }
+      } else {
+        const redirectAfterLoginUrl = redirectToPage
+          ? window.location.href.split('?')[0] + `?page=${ redirectToPage }`
```
v-excelsior, line 19: [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

Can we use it?

Saharij: Of course we can, but why if it already works.

components/bl-social-login-button-component/styles/index.less
```diff
@@ -0,0 +1,149 @@
...
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    height: 50px;
+    width: 100%;
+    line-height: 50px;
+    margin-bottom: 10px;
+    background-color: transparent;
+    box-sizing: border-box;
+	  border-radius: 3px;
```
v-excelsior, line 23: formatting

components/bl-social-login-button-component/styles/index.less
```diff
@@ -0,0 +1,149 @@
...
+    cursor: pointer;
+
+    &:hover {
+      color: #FFFFFF;
```
v-excelsior, line 27: that too must comes from variables

components/bl-social-login-button-component/src/button.js
v-excelsior: Is it stay on the same page(without reloading) if we did not provide redirectPage?
I could assume that the server return a redirect answer on this request because we provide a redirect parameter. 

And if really need a redirect. Do we save all query parameters? As I see we leave only `page`, but what if we need to pass some other parameters(or keep current)?

Maybe @Valodya could help us with correct behavior there.

components/bl-social-login-button-component/src/button.js
```diff
@@ -0,0 +1,52 @@
...
+    <button
+      onClick={ handleClick }
+      className={ cn("social-button", `social-button__${ provider }`, { ["with-icon"]: showButtonIcon }) }>
+      { showButtonIcon &&
+        <div className="social-button__icon-container">
+          { iconsMap[provider] }
+        </div>
+      }
+      <div className="social-button__text-container">
+        <span className="social-button__text">{ `Connect with ${ provider }` }</span>
```
v-excelsior, line 48: `Connect with ` text should comes from properties

v-excelsior (changes requested):

components/bl-social-login-button-component/src/button.js
Valodya: 1) we won't stay on the page because `window.location = await Backendless.UserService.getAuthorizationUrlLink` redirects us to login provider page

2) as far as I know providing the argument `redirectAfterLoginUrl` is required

3) all the query params of the page are already in the PageData so there is no need to keep the current query params, but I assume we have to provide an option to pass extra params which will be applied to a PageData after login (redirection)

components/bl-social-login-button-component/component.json
```diff
@@ -0,0 +1,150 @@
...
+      "hasLogicHandler": true,
+      "handlerId": "iconsVisibilityLogic",
+      "handlerLabel": "Icons Visibility Logic",
+      "dataBinding": true,
+      "settingTooltip": "This handler allows you to control the visibility of the button icon.",
+      "handlerDescription": "This handler allows you to control the visibility of the button icon."
+    },
+    {
+      "type": "checkbox",
+      "name": "googleplus",
```
Valodya, line 63: @v-excelsior @Saharij wouldn't it be better to have a Select to choose a login provider instead of these checkboxes?

the component name is `Social Login Button`, so if I were a customer I would expect the component add only one button

does it make sense?

Valodya: or, rename the component to Social Login Buttons

v-excelsior: I prefer the second variant.

Valodya: ok, @Saharij please rename the directory/styles/docs/etc as well

Valodya: still `bl-social-login-button-component`

PR #36, opened 2022-08-25
MARKET-1327 Accordion

serhmel:
Screenshot_1
Screenshot_2 Screenshot_3

UI Preview:
UI Preview


components/bl-accordion-component/src/index.js
```diff
@@ -0,0 +1,46 @@
...
+import { useItemsState } from './helpers';
+import { Item } from './item';
+
+const { cn } = BackendlessUI.CSSUtils;
+
+export default function AccordionComponent({ component, eventHandlers }) {
+  const { classList, display, accordionData, controlledAccordion, style } = component;
+  const { onClick, onMouseOver, onMouseOut } = eventHandlers;
+
+  const [accordionId, setAccordionId] = useState(() => Math.random().toString(16).slice(-8));
```
Saharij, line 12: We have an ID generator `BackendlessUI.UUID.short()`
`const accordionId = useMemo(() => BackendlessUI.UUID.short(), []);`

components/bl-accordion-component/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Usage
+
+After adding the component to the page, specify the accordion data. This can be done in UI-Builder designer or using codeless logic.
+
+### Styles and Settings
+
+<dl>
+<dt>Accordion Data</dt>
+<dd>Array of objects that define the title and content of the accordion sections. Can be determined by Accordion Data Logic.</dd>
```
v-excelsior, line 15: Could we add a signature for that objects?

components/bl-accordion-component/README.md
```diff
@@ -0,0 +1,58 @@
...
+<dt>Content Padding</dt>
+<dd>Value from this field will be used as the padding for the accordion section content.</dd>
+</dl>
+
+### Event Handlers and Bindable Properties
+
+<dl>
+<dt>Accordion Data Logic</dt>
+<dd>This is the handler for the logic to determine the title and content of the accordion sections. Array of objects returned by the handler will define the data for the accordion.</dd>
+<dt>On Click Event</dt>
```
v-excelsior, line 41: On Item Click Event? Also, from description: "clicks the mouse or taps the title". What if we click on content?

components/bl-accordion-component/src/index.js
```diff
@@ -0,0 +1,46 @@
...
+            key={ index }
+            item={ item }
+            index={ index }
+            component={ component }
+            accordionId={ accordionId }
+            active={ stateStore[index] }
+            onToggle={ () => handleToggle(item, index) }
+          />
+        );
+      })}
```
v-excelsior, line 42:
```suggestion
      { accordionData.map((item, index) => (
        <Item
          key={ index }
          item={ item }
          index={ index }
          component={ component }
          accordionId={ accordionId }
          active={ stateStore[index] }
          onToggle={ () => handleToggle(item, index) }
        />
      )) }
```

components/bl-accordion-component/src/item.js
```diff
@@ -0,0 +1,65 @@
...
+  const contentStyles = {
+    backgroundColor: contentBackgroundColor,
+    color          : contentColor,
+    fontSize       : contentFontSize,
+    padding        : contentPadding,
+  };
+
+  return (
+    <div className={ `item ${ active ? 'active' : '' }` } style={ itemBackgroundColor }>
+      <div
```
v-excelsior, line 42: title component?

components/bl-accordion-component/src/helpers.js
```diff
@@ -0,0 +1,39 @@
+import { useState } from 'react';
+
+export function useItemsState(accordionData, onClick, controlledAccordion) {
+  const [stateStore, setStateStore] = useState(() => {
+    const stateStore = {};
+
+    if (!Array.isArray(accordionData)) {
```
v-excelsior, line 7: What if accordionData = number/undefined/any non array value?

v-excelsior (changes requested):

components/bl-accordion-component/README.md
serhmel: This event is fired only when we open/close some section. This can be done by clicking on the title. So if we click on the content nothing happens. Do you think it is better to move this event to the item?

components/bl-accordion-component/src/helpers.js
serhmel: In this case, the component returns null from index.js:
```
if (!Array.isArray(accordionData)) {
  return stateStore;
}
```

In hook we use it to avoid blocking the page loading through the method forEach.

components/bl-accordion-component/README.md
serhmel: Could you please explain what you mean? 
The default value of the accordion data shows the user how objects should look or that's not what you meant?

components/bl-accordion-component/README.md
v-excelsior: I see two variants

1) Rename it to On Open/Close Item Event

2) Or make it possible to trigger events no matter where we click - on the content or title.

components/bl-accordion-component/src/helpers.js
v-excelsior: I think it's a bad UX, we must validate that and log an error if validation failed.

Because if I understand correctly for now it works like this:

We pass invalid items
Components not rendered
We don't understand why

components/bl-accordion-component/README.md
serhmel: On Toggle Item Event ?

components/bl-accordion-component/README.md
v-excelsior: > The default value of the accordion data shows the user how objects should look or that's not what you meant?

Yes, but that must be documented too. So add a signature at descriptions:

[{title: <string>, content: <string>}]

components/bl-accordion-component/component.json
```diff
@@ -0,0 +1,134 @@
...
+  "category": "Custom Components",
+  "properties": [
+    {
+      "type": "json",
+      "name": "accordionData",
+      "label": "Accordion Data",
+      "showInSettings": true,
+      "hasLogicHandler": true,
+      "handlerId": "accordionDataLogic",
+      "defaultValue": "[\n  {\n    \"title\": \"Collapsible Group Item #1\",\n    \"content\": \"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse malesuada lacus ex, sit amet blandit leo lobortis eget.\"\n  },\n  {\n    \"title\": \"Collapsible Group Item #2\",\n    \"content\": \"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse malesuada lacus ex, sit amet blandit leo lobortis eget.\"\n  },\n  {\n    \"title\": \"Collapsible Group Item #3\",\n    \"content\": \"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse malesuada lacus ex, sit amet blandit leo lobortis eget.\"\n  }\n]",
```
v-excelsior, line 18: Does `\n` comes from UI-Builder?

If no - remove it please.

v-excelsior (changes requested):

components/bl-accordion-component/src/helpers.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    newStore[index] = !stateStore[index];
+    setStateStore(newStore);
+  };
+
+  const updateItemsState = valueCallback => {
+    const newStore = { ...stateStore };
+
+    Object.keys(newStore).forEach(index => newStore[index] = valueCallback(newStore[index]));
```
Valodya, line 34:
```

for (const key in newStore){
  newStore[key] = valueCallback(newStore[key])
}
```

components/bl-accordion-component/src/helpers.js
```diff
@@ -0,0 +1,39 @@
...
+    const newStore = { ...stateStore };
+
+    if (stateStore[index]) {
+      onCloseItem({ item });
+    } else {
+      onOpenItem({ item });
+    }
+
+    if (controlledAccordion) {
+      Object.keys(newStore).forEach(index => newStore[index] = false);
```
Valodya, line 24:
```
for (const key in newStore){
  newStore[key] = false
}
```

PR #106, opened 2022-09-26
BKNDLSS-29876 Leaflet Map

addKiwi:


components/bl-leafletMap-component/src/helpers/coordinates.js
```diff
@@ -9,6 +9,10 @@ export function toCoordinates(coords) {
     return coords.split(',');
   }
 
+  if (typeof coords === 'object' && regEx.test(`${ coords.center.lat },${ coords.center.lng }`)) {
```
v-excelsior, line 12: regEx -> POINT_REGEX

v-excelsior: check for object is enough?

null, [] is objects too

v-excelsior (changes requested):

PR #119, opened 2022-10-04
MARKET-1404 Dropzone component

serhmel:
thumbnail

Screenshot_1

Screenshot_2

UI Preview:
dropzone_ui_preview


components/bl-dropzone-component/component.json
```diff
@@ -0,0 +1,357 @@
...
+      "handlerDescription": "This event is fired when the user clicks the mouse or taps the Clean button.",
+      "contextBlocks": [
+        {
+          "id": "validatedFiles",
+          "label": "Validated Files"
+        }
+      ]
+    }
+  ],
+  "actions": []
```
Saharij, line 356: If actions are not used in the component, then this line should be remove.

components/bl-dropzone-component/styles/index.less
```diff
@@ -0,0 +1,32 @@
+@bl-customComponent-dropzone-backgroundColor: @appBackgroundColor;
+@bl-customComponent-dropzone-textColor: @appTextColor;
+
+@bl-customComponent-dropzone-color: fade(@bl-customComponent-dropzone-textColor, 85%);
+@bl-customComponent-dropzone-labelColor: fade(@bl-customComponent-dropzone-textColor, 60%);
+@bl-customComponent-dropzone-fileName-color: @bl-customComponent-dropzone-textColor;
+@bl-customComponent-dropzone-minWidth: 436px; 
+
+
```
Saharij, line 9: Please remove this extra empty line

v-excelsior:
image

Looks a little messy

v-excelsior:
image

Can we add some space for icons?

components/bl-dropzone-component/src/helpers.js
```diff
@@ -0,0 +1,65 @@
...
+    const updatedFiles = [];
+
+    filesList.forEach(async file => {
+      if (file.uploadStatus === 'uploading') {
+        file.uploadStatus = undefined;
+
+        try {
+          await Backendless.Files.upload(file.file, targetDirectory || '', overwriteFiles);
+
+          file.uploadStatus = 'success';
```
v-excelsior, line 19:
```
const UploadStatus = {
 success: 'succes',
 ...
}
```

v-excelsior (changes requested):

components/bl-dropzone-component/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+The component covers all the main features of the library and provides a way to upload files to a server.
+
+<p align="center">
+  <img src="./thumbnail.png" alt="main thumbnail" width="780"/>
+</p>
+
+## Properties
+
+| Property                | Type                                                                                                                                | Default value                        | Logic                  | Data Binding | UI Setting | Description                                                                              |
```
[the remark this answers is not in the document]
serhmel: It's because of the large width of the table, the formatting is ok.

components/bl-dropzone-component/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+export default function DropzoneComponent({ component, eventHandlers }) {
+  const {
+    display, classList, style, acceptedFileTypes, onDropBehaviour, clickable, footerVisibility, headerVisibility, label,
+    language, maxFiles, maxFileSize, minHeight, maxHeight, viewMode, disableScrollbar, uploadOnDrop, themeColor,
+    cleanButtonVisibility, showInfoLayer, activeItems, preview, borderWidth, borderStyle, borderColor,
+  } = component;
+  const { onDrop, onChangeView, onClean } = eventHandlers;
+
+  const [imageSrc, setImageSrc] = useState(undefined);
```
v-excelsior, line 16: Does it have a difference with
`const [imageSrc, setImageSrc] = useState();` ?

components/bl-dropzone-component/src/helpers.js
```diff
@@ -0,0 +1,97 @@
+import { useState } from 'react';
+
+const UploadStatus = {
+  success  : 'success',
+  uploading: 'uploading',
+  undefined: 'undefined',
```
Valodya, line 6: what does it mean?

components/bl-dropzone-component/src/helpers.js
```diff
@@ -0,0 +1,97 @@
...
+            const directory = targetDirectory?.replace(/\/$/g, '').replace(/(.*)/, '$1/') || '';
+            const path = directory + validFileName;
+
+            await Backendless.Files.upload(file, path, overwriteFiles);
+          }
+
+          item.uploadStatus = UploadStatus.success;
+        } catch (error) {
+          item.uploadStatus = UploadStatus.error;
+          console.error(error.message);
```
Valodya, line 41: would be better to remember the error and pass to a Codeless Logic instead of printing it here

components/bl-dropzone-component/src/helpers.js
```diff
@@ -0,0 +1,97 @@
...
+
+    filesList.forEach(async item => {
+      if (item.uploadStatus === UploadStatus.uploading) {
+        item.uploadStatus = UploadStatus.undefined;
+
+        try {
+          const file = item.file;
+          const hasUploadHandler = available('onUpload');
+
+          if (hasUploadHandler) {
```
Valodya, line 27: please change it to `if (onUpload.hasLogic) {`

components/bl-dropzone-component/src/helpers.js
```diff
@@ -0,0 +1,97 @@
...
+          item.uploadStatus = UploadStatus.success;
+        } catch (error) {
+          item.uploadStatus = UploadStatus.error;
+          console.error(error.message);
+        }
+      }
+
+      updatedFiles.push(item);
+
+      if (updatedFiles.length === filesList.length) {
```
Valodya, line 47:
```
 onChange({ filesList });

for(const fileItem of filesList){
  await ...

  updatedFiles.push(fileItem);
}

setFiles(updatedFiles);
```

components/bl-dropzone-component/src/index.js
```diff
@@ -0,0 +1,90 @@
+import { useState } from 'react';
+
+import { Dropzone, FileItem, FullScreenPreview, VideoPreview } from './lib/dropzone-ui.min';
+import { ensureMeasure, useDropzone } from './helpers';
+
+const { cn } = BackendlessUI.CSSUtils;
+
+export default function DropzoneComponent({ component, eventHandlers }) {
+  const {
```
Valodya, line 9: too many props, please break them into a few logical lines/blocks and then add an empty line in front of `eventHandlers `

```
const { display, classList, style, ... } = component
const { footerVisibility, headerVisibility, ... } = component
const { borderWidth, borderStyle, borderColor, ... } = component
...

 const { onDrop, onChangeView, onClean } = eventHandlers;

```

components/bl-dropzone-component/src/helpers.js
```diff
@@ -0,0 +1,97 @@
...
+      return;
+    }
+
+    const updatedFiles = [];
+
+    files.forEach(file => {
+      if (file.uploadStatus === UploadStatus.success || file.valid === false) {
+        updatedFiles.push(file);
+      } else {
+        updatedFiles.push({ ...file, uploadStatus: UploadStatus.uploading });
```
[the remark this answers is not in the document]
serhmel: By default, the library uses axios to upload files, and the upload status automatically changes according to the response. But in this case we need to change the structure of the server response.
So I decided to use the BL File Upload API to upload files. By setting the upload status to "undefined", I stop using axios. To correctly display the upload status on the page without using axios, we need to set the upload status to "uploading", "success" or "error".

components/bl-dropzone-component/src/helpers.js
serhmel: 'undefined' is a default file upload status. After adding a file to the drop zone, before we click "upload", the file has upload status of "undefined".
Also, if we set the file upload status to "undefined" instead of "uploading",  the file upload will stop.

components/bl-dropzone-component/src/helpers.js
serhmel: If we use this method, the filesList will be updated every time the status changes inside the loop. And this will cause the updateFiles function to work incorrectly and display the wrong file upload status on the page.

So we don't have to wait for the result of the loop. Because in this case the updateFiles function will run only once and the filesList will not change during the cycle.

components/bl-dropzone-component/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+        maxFileSize={ maxFileSize }
+        maxFiles={ maxFiles }
+        label={ label }
+        minHeight={ ensureMeasure(minHeight) }
+        maxHeight={ ensureMeasure(maxHeight) }
+        localization={ language }
+        view={ viewMode === 'unset' ? false : viewMode }
+        footer={ footerVisibility }
+        header={ headerVisibility }
+        url
```
Valodya, line 52: according to the doc https://github.com/dropzone-ui/dropzone-ui-react#props

the `url` property must be a string not boolean

components/bl-dropzone-component/src/helpers.js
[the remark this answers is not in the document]
Valodya: in my point of view, this is a hack which will bring us many problems in the future, can we avoid this?

components/bl-dropzone-component/src/helpers.js
Valodya: hold on, in my version behaviour will be almost the same, the only difference is you won't compare `updatedFiles.length === filesList.length` and iterations will go in a synchronous manner, or did I miss something?

ok, let's get rid of the hack with status and see one more time

PR #147, opened 2022-10-29
BKNDLSS-30263 Pagination

StasivPavlo:


components/bl-pagination-component/src/index.js
```diff
@@ -40,6 +40,12 @@ export default function Pagination({ component, eventHandlers }) {
     setCurrentPage(pageCount);
   };
 
+  useEffect(() => {
+    if (siblingCount < 0) {
+      throw new Error('The sibling count cannot be less than 0');
+    }
+  }, [siblingCount]);
```
v-excelsior, line 47: We have an useMemo that depends on siblingCount. Could we throw this error at the same place(usePagination hook)?

v-excelsior (changes requested):

PR #148, opened 2022-10-29
BKNDLSS-30264 Pagination

StasivPavlo:


components/bl-pagination-component/src/index.js
```diff
@@ -40,6 +40,12 @@ export default function Pagination({ component, eventHandlers }) {
     setCurrentPage(pageCount);
   };
 
+  useEffect(() => {
```
v-excelsior, line 43: We have an useMemo that depends on pageCount. Could we throw this error at the same place(usePagination hook)?

Also:
`The page count must be greater than 0`

v-excelsior (changes requested):

PR #154, opened 2022-11-04
BKNDLSS-30272 Autocomplete

Saharij:


components/bl-autocomplete-component/src/helpers/use-options-placement.js
```diff
@@ -1,12 +1,12 @@
...
 export const useOptionsPlacement = props => {
-  const { optionsRef, autocompleteHeight, setOptionsPlacement } = props;
-  
+  const { optionsContainerRef, autocompleteHeight, setOptionsPlacement } = props;
+
   useLayoutEffect(() => {
     const handleScroll = () => {
       const viewPortHeight = window.innerHeight;
-      const { bottom, height } = optionsRef.current.getBoundingClientRect();
+      const { bottom, height } = optionsContainerRef.current?.getBoundingClientRect();
```
v-excelsior, line 9: `optionsContainerRef.current?.getBoundingClientRect()`

I assume the question mark means - sometimes we haven't ref
In that case `bottom` and `height` becomes undefined and all code below looks broken
`const bottomCoordinate = placement === 'top' ? undefined + autocompleteHeight + undefined  : undefined;`

And at some render iterations we create a subscription with the invalid handler, and remove it. That looks to me like unneeded action. Could we just skip iterations where the ref is not initialized?
```
if(ref.current) {
...
}
```

Do you agree?

components/bl-autocomplete-component/src/index.js
```diff
@@ -45,8 +34,24 @@ export default function AutocompleteComponent({ component, eventHandlers }) {
...
+      className={
+        cn(
+          'bl-customComponent-autocomplete', `bl-customComponent-autocomplete--${variant}`, classList,
+          {
+            'bl-customComponent-autocomplete--disabled': disabled,
+            'has-clear-button': autocompleteValue,
+            'autocomplete-focused': isAutocompleteActive,
+          }
+        )
+      }>
```
v-excelsior, line 54: const classes = {...}

className = { classes }

v-excelsior: Also it could be a function below the component

components/bl-autocomplete-component/src/options/index.js
```diff
@@ -1,29 +1,24 @@
...
 
-  useOptionsPlacement({ optionsRef, autocompleteHeight, setOptionsPlacement });
+  useOptionsPlacement({ optionsContainerRef, autocompleteHeight, setOptionsPlacement });
 
   if(!optionList.length) {
     return (
-      <div className="options">
+      <div className="options" ref={ optionsContainerRef }>
         <div className="option">
           No options
```
v-excelsior, line 23: Add property for that - Empty Options Label

components/bl-autocomplete-component/src/text-field/index.js
```diff
@@ -25,7 +23,7 @@ export const TextField = forwardRef((props, ref) => {
   const { onButtonClearClick, onChange } = eventHandlers;
 
   const inputRef = useRef(null);
-  const autocompleteId = useMemo(() => generateId(), []);
+  const autocompleteId = useMemo(() => BackendlessUI.UUID.short(), []);
```
v-excelsior, line 26: Does React.useId() works here? Not critical, just for information.

v-excelsior (changes requested):

components/bl-autocomplete-component/src/text-field/index.js
Saharij: React.useId() doesn't work here because it appeared in React 18. And we use React 17.

PR #158, opened 2022-11-09
MARKET-1472 Multiple Bar Chart

Saharij:


components/bl-multiple-bar-chart-component/src/validate.js
```diff
@@ -1,21 +1,12 @@
...
-  if (!items[0].hasOwnProperty('objectId')) {
-    return items.map(item => ({
-      ...item,
-      objectId: getId(),
+  if (Array.isArray(items)) {
+    return items.map(({ name, goal, progress }) => ({
+      id: name,
+      name: name,
+      goal: goal > 0 ? goal : 1,
+      progress: progress < 0 ? 0 : progress,
```
v-excelsior, line 7: `Math.max(0, progress)`

components/bl-multiple-bar-chart-component/src/validate.js
```diff
@@ -1,21 +1,12 @@
 export const validate = items => {
-  if (!items) {
-    return [];
-  }
-  
-  if (!items[0].hasOwnProperty('objectId')) {
-    return items.map(item => ({
-      ...item,
-      objectId: getId(),
+  if (Array.isArray(items)) {
```
v-excelsior, line 2:
```
export const validate = items => {
  if (!Array.isArray(items)) {
    return []
  }

  return items.map(({ name, goal, progress }) => ({
    name,
    id: name,
    goal: goal > 0 ? goal : 1,
    progress: Math.max(0,progress),
  }));
};
```

That looks for me more clear and informative

we have corner case with handling:
```
  if (!Array.isArray(items)) {
    return []
  }
  ```
  
  and default behavior:
  
  ```
return items.map(({ name, goal, progress }) => ({
  name,
  id: name,
  goal: goal > 0 ? goal : 1,
  progress: Math.max(0,progress),
}));
  ```

File: components/bl-multiple-bar-chart-component/src/use-chart-data.js [hunk 3]
```diff
@@ -2,7 +2,7 @@ import { useMemo } from 'react';
 
 export const useChartData = (goal, progress) => useMemo(() => {
   const fillPercentage = (progress / (goal / 100)).toFixed(2);
-  const barFillStyle = { width: `${ fillPercentage }%` };
+  const barFillStyle = { width: `${ fillPercentage > 100 ? 100 : fillPercentage }%` };
```
v-excelsior, line 5: Math.min(fillPercentage,100)

components/bl-multiple-bar-chart-component/src/use-chart-data.js
```diff
@@ -2,7 +2,7 @@ import { useMemo } from 'react';
 
 export const useChartData = (goal, progress) => useMemo(() => {
   const fillPercentage = (progress / (goal / 100)).toFixed(2);
```
v-excelsior, line 4: Also it's string

v-excelsior (changes requested):

PR #160, opened 2022-11-10
Simple Modal MARKET-1509

StasivPavlo:


components/bl-simpleModal-component/src/index.js
```diff
@@ -27,7 +25,20 @@ export default function SimpleModal({ component, eventHandlers }) {
...
+    document.body.classList.toggle('active-modal', isOpen);
+  }, [isOpen]);
+
+
+  useEffect(() => {
+    if (!display) {
+      component.closeModal();
+    } else {
+      setIsOpen(true)
+    }
```
v-excelsior, line 38: Avoid negative conditions where it possible

```
    if (display) {
      setIsOpen(true)
    } else {
      component.closeModal();
    }
```

v-excelsior (changes requested):

PR #171, opened 2022-12-05
MARKET-1468 Autocomplete

Saharij:
Screenshot 2022-12-07 at 02 06 09
Screenshot 2022-12-07 at 02 10 44


components/bl-autocomplete-component/src/index.js
```diff
@@ -21,10 +21,20 @@ export default function AutocompleteComponent({ component, eventHandlers }) {
...
+    const isMatched = label => label.toLowerCase().includes(inputValue.toLowerCase());
+
+    if (hasGroup) {
+      const filteredChildren = item.children.filter(child => isMatched(child.label));
+
+      return filteredChildren.length ? [...acc, { ...item, children: filteredChildren }] : acc;
+    }
+
+    return isMatched(item.label) ? [...acc, item] : acc;
+  }, []);
```
v-excelsior, line 37: move to hook or function. Also it could be memoized and maybe we need to add debounce here.

components/bl-autocomplete-component/src/options/index.js
```diff
@@ -30,15 +32,24 @@ export function Options(props) {
...
+              onChange={ onChange }
+            />
+          : <Option
+              key={ item.value }
+              item={ item }
+              setInputValue={ setInputValue }
+              setIsOptionsOpen={ setIsOptionsOpen }
+              setAutocompleteValue={ setAutocompleteValue }
+              onChange={ onChange }
+            />
```
v-excelsior, line 52: const OptionComponent = useMemo(() => hasGroup ? GroupedOptions : Option, [hasGroup])

{ optionsList.map(item => ( <OptionComponent .../>) }

components/bl-autocomplete-component/src/index.js
```diff
@@ -21,10 +21,20 @@ export default function AutocompleteComponent({ component, eventHandlers }) {
...
   }, [options]);
 
+  const hasGroup = optionsList[0]?.hasOwnProperty("groupLabel");
   const autocompleteHeight = autocompleteRef.current?.getBoundingClientRect()?.height;
-  const filteredOptions = optionsList.filter(({ label }) => (
-    label.toLowerCase().includes(inputValue.toLowerCase())
-  ));
+
+  const filteredOptions = optionsList.reduce((acc, item) => {
+    const isMatched = label => label.toLowerCase().includes(inputValue.toLowerCase());
```
v-excelsior, line 28: You create this function on every render `optionList.length` times. Move it out from a component. Also, I could offer another name - compareCaseInsensitive(str1, str2). So you can choose which you prefer more.

components/bl-autocomplete-component/src/helpers/validate.js
```diff
@@ -1,17 +1,7 @@
 export const validate = items => {
-  if (!items) {
+  if (!Array.isArray(items)) {
```
v-excelsior, line 2: Maybe we should add some error logs for this cases? Or at first render it could be empty even if we pass options?

v-excelsior (changes requested):

v-excelsior (changes requested):
This commend have been discussed and edited.

components/bl-autocomplete-component/src/options/index.js
Saharij: agreed, but do we really need useMemo here? reference on `GroupedOptions` or `Option` is static, so we can omit useMemo here, am I correct?

Saharij:
> In general. I don't understand why we add these checkboxes to multiple select component as default. Maybe it should be a variant of view but not a default behavior?
> 
> @Valodya what do you think?

components/bl-autocomplete-component/src/options/index.js
v-excelsior: Yes, my mistake.

v-excelsior (approved):

components/bl-autocomplete-component/src/index.js
```diff
@@ -11,7 +11,7 @@ export default function AutocompleteComponent({ component, eventHandlers }) {
 
   const rootRef = useRef();
   const autocompleteRef = useRef();
-  const [optionsList, setOptionsList] = useState([]);
+  const [optionsList, setOptionsList] = useState(validate(options));
```
Valodya, line 14: `useState(() => validate(options))` in order to do not run it on each render

Valodya: and what if the list is changing in runtime?

for instance:
- we pass a valid list
- then in some time we assign a new `options` to the component

Will we valid the new options?

components/bl-autocomplete-component/src/index.js
```diff
@@ -21,10 +21,10 @@ export default function AutocompleteComponent({ component, eventHandlers }) {
     setOptionsList(validate(options));
   }, [options]);
 
+  const hasGroup = optionsList[0]?.hasOwnProperty("groupLabel");
```
Valodya, line 24: do we really need the `hasOwnProperty ` here?

components/bl-autocomplete-component/src/index.js
Valodya: perhaps `useMemo` fits better for that, what do you think?

components/bl-autocomplete-component/src/index.js
Saharij: How else can we check the object for the presence of the "groupLabel" property?

Valodya: `optionsList[0]?.groupLabel`?

PR #173, opened 2022-12-07
BKNDLSS-30618 Rich Editor

serhmel:
Screenshot_6


components/bl-rich-editor-component/styles/index.less
```diff
@@ -67,6 +68,10 @@
     .ql-picker {
       color: @bl-customComponent-richEditor-textColor;
 
+      &.ql-font {
+        width: @bl-customComponent-richEditor-toolbar-fontPicker-width;
```
v-excelsior, line 72: It couldn't be flex and grown when needed?

serhmel: The default was `108px` since there was no `times new roman` font, so I decided to just increase the width.
As for flex and grown, I could do that, but I don't think it would be a good idea when the toolbar has 2 or 3 lines and it jumps when we change the font.

PR #178, opened 2022-12-12
MARKET-1557 List Component

Fialkaksu:
[https://bugs.backendless.com/browse/MARKET-1557](url)


components/bl-list-component/README.md
```diff
@@ -0,0 +1,55 @@
+# List
+
+List is the component that can be used in Backendless [UI-Builder](https://backendless.com/developers/#ui-builder). It allows you to add a standard list to your application.
+
+<p align="center">
+  <img src="./thumbnail.png" alt="main thumbnail" width="643"/>
```
v-excelsior, line 6: Default width for thumbnail is 780px. Mark will provide production image later, but this row should be changed.

components/bl-list-component/README.md
```diff
@@ -0,0 +1,55 @@
...
+</p>
+
+Additional properties allow customizing the font size, background color, text color, padding and behavior of the component. Refer to the Properties section below.
```
v-excelsior, line 9: Redundant description

components/bl-list-component/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Configuration
+
+The configuration can be done in the UI Builder or using the Codeless Logic. You need to select the type of list (ordered/unordered) and specify the data that will be displayed inside the list.
```
v-excelsior, line 13: Redundant description

You will add info about list types into at start of readme. Also helpful link to mdn will be good too.

components/bl-list-component/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Properties
+
+| Property                 | Type       | Default value | Logic              | Data Binding | UI Setting | Description                                                        |
+|--------------------------|------------|---------------|--------------------|--------------|------------|-------------------------------------------------------------------|
+| Type                     | *Select*   | `"Unordered"` | Type Logic         | NO           | YES        | Controls the type of list(ul/ol).                                                               |
```
v-excelsior, line 19: Make as [here](https://github.com/Backendless/ui-builder-library/tree/main/components/bl-multiple-bar-chart-component) (Grid Marks property is example)

components/bl-list-component/README.md
```diff
@@ -0,0 +1,55 @@
...
+| List Items               | *JSON*     |               | ListItems Logic    | YES          | NO         | Specifies a JSON array containing data for the list items. Watch [Codeless Examples] #codeless-examples). |
+| Background               | *Color*    |               |                    | NO           | YES        | Controls the background color of the main block.                                                        |
+| Width                    | *Text*     |               |                    | NO           | YES        | Controls the width of the main block.                                                             |
+| Color                    | *Color*    |               |                    | NO           | YES        | Controls the color of the list items.                                                             |
+| Font Size                | *Text*     |               |                    | NO           | YES        | Controls the font size of the list items.                                                             |
```
v-excelsior, line 24: Fix table markdown with https://www.tablesgenerator.com/markdown_tables

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,19 @@
+import { List } from './list';
+
+export default function ListComponent ({
+  component, eventHandlers
+}) {
```
v-excelsior, line 5: In one line 

```suggestion
export default function ListComponent ({ component, eventHandlers }) {
```

Also, did you setup eslint and editorconfig in your IDE?
https://github.com/Backendless/eslint-config
https://github.com/v-excelsior/editor-config/blob/main/.editorconfig

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,19 @@
...
+
+  const { cn } = BackendlessUI.CSSUtils;
```
v-excelsior, line 7: That should be moved out from the component. You don't need to create cn every render

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,19 @@
...
+  const { style, classList, width, backgroundColor } = component;
+
+  style.width = width;
+  style.backgroundColor = backgroundColor;
+
```
v-excelsior, line 12:
```suggestion
const styles = useMemo(()=> ({...style, width, backgroundColor}), [style, width, backgroundColor])
```

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <div className={ cn('bl-customComponent-list', classList) } style={ style }>
+      <List component={component}
+            eventHandlers={eventHandlers}/>
```
v-excelsior, line 16:
```suggestion
      <List component={component} eventHandlers={eventHandlers}/>
```

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
+import { List } from './list';
+
+export function ListContent({
```
v-excelsior, line 3: in one line

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
...
+
+ const getItems = (items) => {
+    return (items || []).map((item, i) => {
+      if (item.child && item.child.length !== 0) {
+          return (
+            <li key={i.toString()}
+              className="list__item"
+              style = {itemStyles}
+              onClick={(e) => {
+                e.stopPropagation();
```
v-excelsior, line 25: Do we realy need stop propagation here?

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
...
+
+  const onItemClick = (key) => {
+    eventHandlers.onClickListItem({item: key});
+  }
+
+ const getItems = (items) => {
+    return (items || []).map((item, i) => {
+      if (item.child && item.child.length !== 0) {
+          return (
+            <li key={i.toString()}
```
v-excelsior, line 21: Could be without `toString`

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
...
+              <List component={component}
+                    child={item.child}
+                    eventHandlers={eventHandlers}/>
+            </li>
+          )
+      } else {
+        return (
+          <li key={i.toString()}
+            className="list__item"
+            style = {itemStyles}
```
v-excelsior, line 39: setup eslint and editorconfig

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
...
+
+export function ListContent({
+  component, eventHandlers, child}) {
+
+  const { color, fontSize, listItems, type} = component;
+
+  const itemStyles = {
+    color   : color,
+    fontSize: fontSize
+  };
```
v-excelsior, line 11: redundant variable

Also 

`{color, fontSile}` works good too

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
...
+
+  const onItemClick = (key) => {
+    eventHandlers.onClickListItem({item: key});
+  }
+
+ const getItems = (items) => {
```
v-excelsior, line 17: That was named a Render Function. In general, we wan't create them - because we write on react and there are preferable component approach.

You made it almost right - but there is a mistake in entities.

We must have an two components:

ListItem - that render an ordinary item (li) and creates a sibling List if it has children.
List - that render a collection of items

v-excelsior: Also, pseudo code for that looks like:

```
function List({ children }) {
  return (
    <ul>
      { children.map(child => {
        return (
          <Item { ...child }/>
        )
      }) }
    </ul>
  )
}

function Item(itemProps) {
  return (
    <>
      <li>{ itemProps.text } </li>
      { itemProps.children && <List children={ itemProps.children }/> }
    </>
  )
}
```

components/bl-list-component/README.md
```diff
@@ -0,0 +1,55 @@
...
+## Configuration
+
+The configuration can be done in the UI Builder or using the Codeless Logic. You need to select the type of list (ordered/unordered) and specify the data that will be displayed inside the list.
+
+## Properties
+
+| Property                 | Type       | Default value | Logic              | Data Binding | UI Setting | Description                                                        |
+|--------------------------|------------|---------------|--------------------|--------------|------------|-------------------------------------------------------------------|
+| Type                     | *Select*   | `"Unordered"` | Type Logic         | NO           | YES        | Controls the type of list(ul/ol).                                                               |
+| List Items               | *JSON*     |               | ListItems Logic    | YES          | NO         | Specifies a JSON array containing data for the list items. Watch [Codeless Examples] #codeless-examples). |
```
v-excelsior, line 20: Describe signature

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
...
+    return (items || []).map((item, i) => {
+      if (item.child && item.child.length !== 0) {
```
v-excelsior, line 19: `child` should be plural - `children`

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,30 @@
+import { ListContent } from './list-content';
+
+export function List ({
+  component, eventHandlers, child
+}) {
+
+  const listType = (type) => {
+    if(type && type === 'ol') {
```
v-excelsior, line 8: Create map for enums

const ListTypes = {
  ORDERED: 'ol',
  UNORDERED: 'ul'
}

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
...
+        return (
+          <li key={i.toString()}
+            className="list__item"
+            style = {itemStyles}
+            onClick={(e) => {
+              e.stopPropagation();
+              onItemClick(item.item);
+            }}
+          >
+            {item.item}
```
v-excelsior, line 45: item.content

components/bl-list-component/src/list-content.js
v-excelsior, line 19: this cons
```suggestion
    return (items || []).map((item, i) => {
      if (item.child && item.child.length !== 0) {
```

this construction should hone after refactoring to List and ListItem component.

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
...
+          </li>
+        )
+      }
```
v-excelsior, line 48: That should gone after other changes, but you was able short it twice

```suggestion
          return (
            <li key={i}
              className="list__item"
              style={itemStyles}
              onClick={(e) => onItemClick(e, item) }
            >
              { item.content }
              { item.child && (
                  <List component={component}
                    child={item.child}
                    eventHandlers={eventHandlers}
                  />
               ) }
            </li>
          )
      }
```

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+  }
+
+  return (getItems(child || listItems))
```
v-excelsior, line 52: about returning result of render function. Return common jsx.

If code forces you to write render function - try to split parts of it into components.

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,30 @@
+import { ListContent } from './list-content';
+
+export function List ({
```
v-excelsior, line 3: one line

v-excelsior (changes requested):

components/bl-list-component/component.json
```diff
@@ -0,0 +1,88 @@
+{
+  "id": "c_f95c74e56529269bdafe747715db7bbe",
+  "name": "List",
+  "description": "The List component allows the user to add a standard list to an application.",
+  "showInToolbox": true,
+  "faIcon": "pencil-ruler",
```
v-excelsior, line 6: change icon

components/bl-list-component/component.json
```diff
@@ -0,0 +1,88 @@
...
+        {
+          "value": "ul",
+          "label": "Unordered"
+        },
+        {
+          "value": "ol",
+          "label": "Ordered"
+        }
+      ],
+      "handlerLabel": "Type Logic"
```
v-excelsior, line 29: add handler descriptions and setting tooltips everywhere where it possible

v-excelsior (changes requested):

components/bl-list-component/src/list.js
v-excelsior: Also read this article https://www.robinwieruch.de/react-as-prop/

briefly:

```
<List as="ul" ...>
```

and in list component:
```
const Component = as

return (<Component ... />)
```

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,18 @@
+import { useMemo } from 'react'
+
+import { List } from './list'
+
+const { cn } = BackendlessUI.CSSUtils
+
+export default function ListComponent ({ component, eventHandlers }) {
+
```
v-excelsior, line 8: No need in this empty line

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,18 @@
...
+  const { style, classList, width, backgroundColor } = component
+
+  const styles = useMemo(() => ({ style, width, backgroundColor }), [style, width, backgroundColor])
```
v-excelsior, line 11: const styles = { ...style, width, backgroundColor }. Also we could do it props without creating variable.

There are no need in useMemo as we haven't any calculations

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,16 @@
+import { ListContent } from './list-content'
+
+export function List ({ component, eventHandlers, children, as=component.type }) {
```
v-excelsior, line 3: Why we need  `as=component.type`?

v-excelsior, line 3: `<List as={ component.type } component={ component } eventHandlers={ eventHandlers }/>`

As i see we don't pass children here.

But in another place:

```
<List
          component={ component }
          children={ item.children }
          eventHandlers={ eventHandlers }
        />
```

Could we add an props 'items' - and only this props can accept collection of objects to render?

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,29 @@
+import { List } from './list'
+
+export function ListContent({ component, eventHandlers, children }) {
+
+  const { color, fontSize, listItems } = component
```
v-excelsior, line 5: Use different props, don't pass full component object everywhere

components/bl-list-component/styles/index.less
```diff
@@ -0,0 +1,25 @@
+@bl-customComponent-list-theme: @themePrimary;
+@bl-customComponent-list-themeTextColor: @appTextColor;
+
+@bl-customComponent-list-item-hover-background-color: if(@isLightTheme, rgba(0, 0, 0, 0.04), rgba(255, 255, 255, 0.04));
+
+@bl-customComponent-list-item-size: 20px;
```
v-excelsior, line 6: list-item-size -> list-item-fontSize

v-excelsior (changes requested):

components/bl-list-component/styles/index.less
Fialkaksu: this is a general parameter for size control, below is used to set the margins/paddings. Perhaps, in this case, it would be better to rename to indent or leave it as it is.

components/bl-list-component/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+<p align="center">
+  <img src="./thumbnail.png" alt="main thumbnail" width="780"/>
+</p>
+
+## Properties
+
+| Property   | Type                                          | Default value | Logic           | Data Binding | UI Setting | Description                                                                                      |
+|------------|-----------------------------------------------|---------------|-----------------|--------------|------------|--------------------------------------------------------------------------------------------------|
+| Type       | *Select* <br> `'Unordered'` <br> `'Ordered'`  | `Unordered`   | Type Logic      | NO           | YES        | Controls the type of list(ul/ol).                                                                |
```
v-excelsior, line 14: There should be real values, not labels
Unordered -> ul
Ordered - ol

v-excelsior: Check the example
https://github.com/Backendless/ui-builder-library/blob/main/components/bl-leafletMap-component/README.md

components/bl-list-component/component.json
```diff
@@ -0,0 +1,95 @@
...
+  "name": "List",
+  "description": "The List component allows the user to add a standard list to an application.",
+  "showInToolbox": true,
+  "faIcon": "list",
+  "mainJS": "dist/index.js",
+  "type": "custom",
+  "category": "Custom Components",
+  "properties": [
+    {
+      "type": "select",
```
v-excelsior, line 12: Need to add handler descriptions

Check example https://github.com/Backendless/ui-builder-library/blob/main/components/bl-leafletMap-component/component.json

components/bl-list-component/component.json
```diff
@@ -0,0 +1,95 @@
...
...UWvu0CXtE19cRd+nTmTh8lD0gI7lUr6LOxAYJxf0py2x4f1fM3tbfM76yDJXv4u1M7DqPUo4tb/nuUIzTh15UvR8nXO9P9rJ70YrvTimtz7UNsSXrzHFxWdE4FCv/wHedtKMT/bCUdNt0ez4UJ3oWcjCvL6dxzr7ViqXlOLG6viVG6p1hxdgcFCuPhkmRtHyn9pHbV6j1kYN5VWXso3qdMx/Rth0rA5loJmcsB4Wj183LJKlXdp3VQQBZx9/Jthvc9pDsLMqKm6xp2eOhS0fVpdDg1duuKurGhSdF0lImGtO0zNRAdSwO1BkBu25O/Oo+KVLrsxRnz8fpg32Sh/pE6n2XY4jSvNq0u0TXY6uiaNk36k/CsTk4/lBqL6OiX1S6iMNXFEkRcGnsZTHu5VJft+R+on6fbgcAAF8/kiLgMpnLJTYL15B/ncxR+a7PfAEAAFwykiLgUqlLPwYNbqT/50oeZNHvdxoAAICrgaQIuGwfZvqJMTsvvubzK8lTbEKPAwcAAPiakBQBfwP9RLKbk5pP1/nn0Y9B/YrXHwAAwEVSBAAAAKDXSIoAAAAA9BpJEQAAAIBeIykCAAAA0GskRQAAAAB6jaQIAAAAQK+RFAEAAADoNZIiAAAAAL1GUgQAAACg10iKAAAAAPQaSREAAACAXiMpAgAAANBrJEUAAAAAeo2kCAAAAECvkRQBAAAA6DWSIgAAAAC9RlIEAAAAoNdIigAAAAD0GkkRAAAAgF4jKQIAAADQayRFAAAAAHqNpAgAAABAr5EUAQAAAOg1kiIAAAAAvUZSBAAAAKDXSIoAAAAA9BpJEQAAAIBeIykCAAAA0GskRQAAAAB6jaQIAAAAQK+RFAEAAADoNZIiAAAAAL1GUgQAAACg10iKAAAAAPQaSREAAACAXiMpAgAAANBrJEUAAAAAeo2kCAAAAECvkRQBAAAA6DWSIgAAAAC9RlIEAAAAoNdIigAAAAD0GkkRAAAAgF4jKQIAAADQayRFAAAAAHqNpAgAAABAr5EUAQAAAOg1kiIAAAAAPSbE/wP77WLQJEzlqgAAAABJRU5ErkJggg==\" data-module-type=\"system\" data-module-id=\"image\" data-display data-uid=\"9bfbec347a9aa9e67ac32eff75a98333\"/></div>"
```
v-excelsior, line 94: Why we use image for that? Could we create an markup which looks like list?

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,28 @@
...
+    type: type && type === "ol" ? "ol" : "ul",
+    color: color,
+    fontSize: fontSize,
+    onClickHandler: onClickHandler,
+  };
+
+  return (
+    <div
+      className={cn("bl-customComponent-list", classList)}
+      style={{ width: width, backgroundColor: backgroundColor }}
```
v-excelsior, line 23: `{ width: width, backgroundColor: backgroundColor }` is same as `{ width, backgroundColor }`

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,15 @@
...
+export function ListContent({ props, items }) {
+  return (items || []).map((item, i) => (
+    <li
+      key={i}
+      className="list__item"
+      style={{ color: props.color, fontSize: props.fontSize }}
+      onClick={(e) => props.onClickHandler(e, item)}
+    >
+      {item.content}
+      {item.children && <List props={props} items={item.children} />}
```
v-excelsior, line 12: Avoid using `props` as name for one of props

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,15 @@
+import { List } from "./list";
+
+export function ListContent({ props, items }) {
+  return (items || []).map((item, i) => (
+    <li
+      key={i}
+      className="list__item"
+      style={{ color: props.color, fontSize: props.fontSize }}
+      onClick={(e) => props.onClickHandler(e, item)}
```
v-excelsior, line 9: Props like `color/fontSize/onItemClick/list type` is stable and everywhere the same, use a react context and pass all of them through it

components/bl-list-component/styles/index.less
v-excelsior: So we must create variables for each of this properties

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,11 @@
+import { ListContent } from "./list-content";
+
+export function List({ props, items }) {
+  const ComponentType = props.type;
+
+  return (
+    <ComponentType className="list">
+      <ListContent props={props} items={items} />
```
v-excelsior, line 8: There is no need for List Content component

`{ items.map(item => (<Item .../>)) }`

components/bl-list-component/src/list-content.js
```diff
@@ -0,0 +1,15 @@
+import { List } from "./list";
+
+export function ListContent({ props, items }) {
+  return (items || []).map((item, i) => (
+    <li
+      key={i}
+      className="list__item"
+      style={{ color: props.color, fontSize: props.fontSize }}
```
v-excelsior, line 8: Use destruction in similar cases

const { color, fontSize } = props

v-excelsior (changes requested):

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,28 @@
+import { List } from "./list";
+
+const { cn } = BackendlessUI.CSSUtils;
+
+export default function ListComponent({ component, eventHandlers }) {
+  const { classList, width, backgroundColor, type, listItems, color, fontSize } = component;
+
+  const onClickHandler = (e, key) => {
+    e.stopPropagation();
```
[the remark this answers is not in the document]
Fialkaksu: Not handled, event bubbles.

components/bl-list-component/README.md
```diff
@@ -0,0 +1,46 @@
...
+|------------|----------------------------|---------------|-----------------|--------------|------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Type       | *Select* <br> [`ul`, `ol`] | `ul`          | Type Logic      | NO           | YES        | Controls the type of list(ul/ol).                                                                                                                                                                   |
+| Items List | *JSON*                     | `[]`          | ListItems Logic | YES          | NO         | Specifies a JSON array containing data of the list items. Watch [Codeless Examples] (#Examples). Signature of items list: [ <Item> ]. Signature of item: { content: String, children: [ <Item> ] }. |
```
v-excelsior, line 15: What if we add property type to the item signature? That could help us for creating an ordered lists inside an unordered and vice versa

v-excelsior (changes requested):

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,23 @@
+export function List({ options, items }) {
+  const ComponentType = options.type;
+
+  return (
+    <ComponentType className="list">
+      {items?.map((item, i) => (
```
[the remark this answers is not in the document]
v-excelsior: In other way.

```
{ listItem.children && <List items={listItem.children} options={options} type={listItem.typeOfList} /> }
```

and here 

```
  { items.map((item, i) => ( ...
```

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,23 @@
...
+}
+
+function Item(props) {
+  const { options, listItem, key } = props;
+  const { style, onClickHandler } = options;
+
+  return (
+    <li key={key} className="list__item" style={style} onClick={(e) => onClickHandler(e, listItem)}>
+      {listItem.content}
+      <List items={listItem.children} options={options} type={listItem.typeOfList} />
```
v-excelsior, line 20: I think you nave not applied editor config

That should look in this way.

```suggestion
     <List items={ listItem.children } options={ options } type={ listItem.typeOfList }/>
```

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const options = {
+    type: type && type === "ol" ? "ol" : "ul",
+    style: {
+      color,
+      fontSize,
+      padding,
+    },
+    onClickHandler: onClickHandler,
+  };
```
v-excelsior, line 23:
```suggestion
  const options = {
    type: type === "ol" ? "ol" : "ul",
    style: {      color,      fontSize,      padding    },
    onClickHandler,
  };
```

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const listRef = useRef();
+
+  component.el = listRef.current;
```
v-excelsior, line 27: https://github.com/Backendless/ui-builder-library/pull/195#discussion_r1084203591

>I'm not sure this is the correct way
let's imaging this was rendered only once and the animationOnScrollRef.current before first render will be null and as result it won't work
I assume we need to wrap it with useEffect

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,34 @@
+import { useRef } from "react";
+
+import { List } from "./list";
+
+const { cn } = BackendlessUI.CSSUtils;
+
+export default function ListComponent({ component, eventHandlers }) {
+  const { classList, width, backgroundColor, type, listItems, color, fontSize, padding } = component;
+
+  const onClickHandler = (e, key) => {
```
v-excelsior, line 10: 'key' -> 'item'

eventHandlers.onClickListItem({ item });

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,34 @@
+import { useRef } from "react";
+
+import { List } from "./list";
+
+const { cn } = BackendlessUI.CSSUtils;
+
+export default function ListComponent({ component, eventHandlers }) {
+  const { classList, width, backgroundColor, type, listItems, color, fontSize, padding } = component;
```
v-excelsior, line 8: Here is type, but in item signature is typeOfList - that should be the same. Also, don;t forget to change XML example.

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,23 @@
+export function List({ options, items, type }) {
+  const ComponentType = type || options.type;
```
v-excelsior, line 2:
```suggestion
  const ListElement = type || options.type;
```

components/bl-list-component/styles/index.less
```diff
@@ -0,0 +1,19 @@
...
+.bl-customComponent-list {
+  color: @bl-customComponent-list-themeTextColor;
+  user-select: none;
+
+  .list {
+
+    &__item {
+      font-size: @bl-customComponent-list-item-fontSize;
+      padding: @bl-customComponent-list-item-fontSize / 2;
+      line-height: 1.3;
```
v-excelsior, line 15: variable for that

components/bl-list-component/styles/index.less
```diff
@@ -0,0 +1,19 @@
...
+
+.bl-customComponent-list {
+  color: @bl-customComponent-list-themeTextColor;
+  user-select: none;
+
+  .list {
+
+    &__item {
+      font-size: @bl-customComponent-list-item-fontSize;
+      padding: @bl-customComponent-list-item-fontSize / 2;
```
v-excelsior, line 14: variable for that

components/bl-list-component/styles/index.less
```diff
@@ -0,0 +1,19 @@
+@bl-customComponent-list-theme: @themePrimary;
+@bl-customComponent-list-themeTextColor: @appTextColor;
+
+@bl-customComponent-list-item-fontSize: 1rem;
+
+.bl-customComponent-list {
+  color: @bl-customComponent-list-themeTextColor;
+  user-select: none;
```
v-excelsior, line 8: variable for that

Also, I do not quite understand why we can't user-select for list content. It's disabling copying text from the list, which is pretty annoying for me personally.

I believe we could do remove this line, or leave with default value and create variable for that(which is better). Or - we could create an property for that.

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,23 @@
...
+
+  return (
+    <ComponentType className="list">
+      {items?.map((item, i) => (
+        <Item listItem={item} key={i} options={options} />
```
v-excelsior, line 7: We don't need to pass full item into Item component using item object.
Create different props for content, children and so on.

Also, at onClick handler pass different context block for content and children

v-excelsior (changes requested):

components/bl-list-component/src/index.js
Fialkaksu: It's a correct names. 'type' is a global type of a list, if we didn't have a type as 'typeOfList' parameter for a child, we get it as a global type.

v-excelsior: That not answering why we have two entities such as type and typeOfList that do absolutely the same thing.

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,38 @@
...
+  const onClickHandler = (e, itemContent, itemChildren) => {
+    e.stopPropagation();
+    eventHandlers.onClickListItem({ itemContent, itemChildren });
+  };
+
+  const options = {
+    type: type && type === "ol" ? "ol" : "ul",
+    style: { color, fontSize, padding },
+    onClickHandler: onClickHandler,
+  };
```
v-excelsior, line 21: https://github.com/Backendless/ui-builder-library/pull/178#discussion_r1088733002

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,38 @@
...
+const { cn } = BackendlessUI.CSSUtils;
+
+export default function ListComponent({ component, eventHandlers }) {
+  const { classList, display, width, backgroundColor, type, listItems, color, fontSize, padding } = component;
+
+  const items = listItems || [];
+
+  const onClickHandler = (e, itemContent, itemChildren) => {
+    e.stopPropagation();
+    eventHandlers.onClickListItem({ itemContent, itemChildren });
```
v-excelsior, line 14: onClickListItem - onItemClick

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,38 @@
...
+    type: type && type === "ol" ? "ol" : "ul",
+    style: { color, fontSize, padding },
+    onClickHandler: onClickHandler,
+  };
+
+  const listRef = useRef();
+
+  useEffect(() => {
+    component.el = listRef.current;
+  }, []);
```
v-excelsior, line 27: listRef.current shouldn't be in ref?

components/bl-list-component/component.json
```diff
@@ -0,0 +1,110 @@
...
+    },
+    {
+      "type": "json",
+      "name": "listItems",
+      "showInSettings": false,
+      "hasLogicHandler": true,
+      "handlerId": "listItemsLogic",
+      "handlerLabel": "List Items Logic",
+      "dataBinding": true,
+      "defaultValue": "",
```
v-excelsior, line 40: add default value

v-excelsior (changes requested):

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,25 @@
...
+    </ListElement>
+  );
+}
+
+function Item(props) {
+  const { options, key, content, children, type } = props;
+  const { style, onClickHandler } = options;
+
+  return (
+    <li key={ key } className="list__item" style={ style } onClick={ (e) => onClickHandler(e, content, children) }>
```
Valodya, line 20: have you checked what `children` look like?

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,30 @@
...
+  const items = listItems || [];
+
+  const onClickHandler = (e, itemContent, itemChildren) => {
+    e.stopPropagation();
+    eventHandlers.onItemClick({ itemContent, itemChildren });
+  };
+
+  const options = {
+    type: type === "ol" ? "ol" : "ul",
+    style: { color, fontSize, padding },
```
Valodya, line 17: what do we apply this ' style` to? for List or ListItem?

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,30 @@
+import { List } from "./list";
+
+const { cn } = BackendlessUI.CSSUtils;
+
+export default function ListComponent({ component, elRef, eventHandlers }) {
+  const { classList, display, width, backgroundColor, type, listItems, color, fontSize, padding } = component;
+
+  const items = listItems || [];
+
+  const onClickHandler = (e, itemContent, itemChildren) => {
```
Valodya, line 10: onClickHandler vs onItemClick

what's true?

components/bl-list-component/src/index.js
```diff
@@ -0,0 +1,30 @@
...
+    e.stopPropagation();
+    eventHandlers.onItemClick({ itemContent, itemChildren });
+  };
+
+  const options = {
+    type: type === "ol" ? "ol" : "ul",
```
Valodya, line 16: why don't we just pass the `type` to the List? and let it decide what to do with the type

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,25 @@
...
+
+        return <Item key={ i } content={ content } children={ children } type={ type } options={ options } />;
+      }) }
+    </ListElement>
+  );
+}
+
+function Item(props) {
+  const { options, key, content, children, type } = props;
+  const { style, onClickHandler } = options;
```
Valodya, line 17: can we avoid using the `options`?

components/bl-list-component/component.json
```diff
@@ -0,0 +1,110 @@
...
+          "label": "Ordered"
+        }
+      ],
+      "handlerLabel": "Type Logic",
+      "settingTooltip": "Select a list type"
+    },
+    {
+      "type": "json",
+      "name": "listItems",
+      "showInSettings": false,
```
Valodya, line 35: why do we disallow specifying it in Settings?

components/bl-list-component/README.md
```diff
@@ -0,0 +1,62 @@
...
+| Items List <br> `listItems`       | *JSON*                                       | `[]`           | ListItems Logic | YES          | NO         | Specifies a JSON array containing data of the list items. Watch [Codeless Examples](#Examples). Signature of items list: `[ <Item> ]`. Signature of item: `{ content: String, children: [ <Item> ], type: String }`. |
+| Background <br> `backgroundColor` | *Color*                                      |                |                 | NO           | YES        | Controls the background color of the main block.                                                                                                                                                                     |
+| Width <br> `width`                | *Text*                                       |                |                 | NO           | YES        | Controls the width of the main block.                                                                                                                                                                                |
```
Valodya, line 17: I assume the property is redundant

components/bl-list-component/README.md
```diff
@@ -0,0 +1,62 @@
...
+| Color <br> `color`                | *Color*                                      |                |                 | NO           | YES        | Controls the color of the list items.                                                                                                                                                                                |
+| Font Size <br> `fontSize`         | *Text*                                       |                |                 | NO           | YES        | Controls the font size of the list items.                                                                                                                                                                            |
```
Valodya, line 19: I assume the property is redundant

components/bl-list-component/styles/index.less
```diff
@@ -0,0 +1,24 @@
...
+
+@bl-customComponent-list-item-cursor: pointer;
+
+.bl-customComponent-list {
+  color: @bl-customComponent-list-themeTextColor;
+
+  .list {
+    margin: @bl-customComponent-list-margin;
+
+    &__item {
```
Valodya, line 17: I would recommend you avoid deep nesting, in the CSS it will look like:

```
.bl-customComponent-list .list {...}
.bl-customComponent-list .list .list__item {...}
```

what about:

```
.bl-customComponent-list {

  .list{} 

  .list__item{}
}
```

components/bl-list-component/src/index.js
Fialkaksu: Apply to the ListItem

components/bl-list-component/src/index.js
Fialkaksu: Changed to onItemClick

components/bl-list-component/src/index.js
Fialkaksu: The type initially passed from the user is applied to the external list, and then, if there are nested lists and the type is specified, then it is applied to the nested list, if not passed, then the external one is applied. This is the specifics of this component, in the course of development we decided to add such behavior.

components/bl-list-component/src/list.js
```diff
@@ -0,0 +1,25 @@
+export function List({ options, items, type }) {
+  const ListElement = type || options.type;
+
+  return (
+    <ListElement className="list">
+      { items.map((item, i) => {
+        const { content, children, type } = item;
+
+        return <Item key={ i } content={ content } children={ children } type={ type } options={ options } />;
```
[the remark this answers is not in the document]
Fialkaksu: It was Dima's recommendation. Need to change?

components/bl-list-component/src/list.js
Fialkaksu: I think we can't, this parameter is needed both for the list and for the item inside.

PR #182, opened 2022-12-16
MARKET-1463 Mention

addKiwi:
image
image


components/bl-mention-component/styles/primereact.less
```diff
@@ -0,0 +1 @@
...lay:none}.p-treetable-scrollable-wrapper{position:relative}.p-treetable-scrollable-header,.p-treetable-scrollable-footer{overflow:hidden;border:0 none}.p-treetable-scrollable-body{overflow:auto;position:relative}.p-treetable-virtual-table{position:absolute}.p-treetable-frozen-view .p-treetable-scrollable-body{overflow:hidden}.p-treetable-unfrozen-view{position:absolute;top:0;left:0}.p-treetable-reorder-indicator-up,.p-treetable-reorder-indicator-down{position:absolute;display:none}.p-treetable .p-treetable-loading-overlay{position:absolute;display:flex;align-items:center;justify-content:center;z-index:1}.p-virtualscroller{position:relative;overflow:auto;contain:strict;transform:translateZ(0);will-change:scroll-position;outline:0 none}.p-virtualscroller-content{position:absolute;top:0;left:0;min-height:100%;min-width:100%;will-change:transform}.p-virtualscroller-spacer{position:absolute;top:0;left:0;height:1px;width:1px;transform-origin:0 0;pointer-events:none}.p-virtualscroller-loader{position:sticky;top:0;left:0;width:100%;height:100%}.p-virtualscroller-loader.p-component-overlay{display:flex;align-items:center;justify-content:center}.p-virtualscroller-loading-icon{font-size:2rem}
```
serhmel, line 1: I'm not sure that the Mention component needs the styles of all Primereact components.

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
+import { useEffect, useState } from 'react';
+
+import primereact from './lib/core';
+
+const { Mention } = primereact.mention;
+const { cn } = BackendlessUI.CSSUtils;
+
+const DEFAULT_FIELD_BLACK_LIST = ['created', '___class', 'ownerId', 'updated', 'objectId'];
```
v-excelsior, line 8: BLACK_LIST -> BLACKLIST

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+  } = component;
+  const { onChange, onFocus, onBlur, onShow, onHide } = eventHandlers;
+
+  const [processedSuggestions, setProcessedSuggestions] = useState([]);
+  const triggers = useTriggers(trigger);
+  const [fieldBlackList, setFieldBlackList] = useState(DEFAULT_FIELD_BLACK_LIST);
+
+  useEffect(() => {
+    if (hideField?.length) {
+      const hideValues = hideField.split(',').map(el => el.trim());
```
v-excelsior, line 23: hiddenFields

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+      setFieldBlackList(prev => [...prev, ...hideValues]);
```
v-excelsior, line 24: If for first render giddenFields = ['1'] , and for second  ['2']. Why we merge them?

v-excelsior, line 24: setFieldsBlacklist

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+  const itemTemplate = (suggestion, options) => {
+    const trigger = options.trigger;
+    const fields = [];
+
+    for (const field in suggestion) {
+      if (!fieldBlackList.includes(field)) {
+        fields.push(field);
+      }
+    }
```
v-excelsior, line 59: O^2 complexity. Blacklist could be a map.

v-excelsior: Also

`const shownFields = suggestions.filter(suggestion => !fieldsBlacklistMap[suggestion])`

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+    if (fields.indexOf('img') > 0) {
+      const index = fields.indexOf('img');
+      [fields[0], fields[index]] = [fields[index], fields[0]];
+    }
```
v-excelsior, line 64: I don't clearly understand what happens here.

Fields have the same structure as suggestions
suggestion has the next structure:

{ field, name, img }

So I'm a little messed up here.

Also, such parts of code should be moved to an external function, to make it more readable.

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+    const availableTrigger = triggers.some(availableTrigger => availableTrigger === trigger);
```
v-excelsior, line 66: availableTrigger -  means we have an entity that is available. But method some returns boolean value.
So

If it realy should be 'availableTrigger' - use find method(for example)
But if it should be boolean - rename it to hasAvailableTrigger / hasTrigger / isTriggerAvailable

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+    if (availableTrigger) {
+      return (
+        <div className="content">
+          { fields.map(field => {
+            if (field === 'img' && suggestion.img) {
```
v-excelsior, line 72: https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings

Use constants or objects for magic strings and numbers

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+      } else {
+        finalSuggestions = orderedSuggestions
+          .filter(suggestion => suggestion.field.toLowerCase().startsWith(query.toLowerCase()));
+      }
+
+      setProcessedSuggestions(finalSuggestions);
+    }
+  };
+
+  const itemTemplate = (suggestion, options) => {
```
v-excelsior, line 51: itemTemplate - ItemRenderer
 
 Also, may there needed useCallback?

components/bl-mention-component/src/index.js
v-excelsior, line 23: `.split(',').map(el => el.trim())` - used twice, create helper function with name stringToList(or something like that)

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+  const [fieldBlackList, setFieldBlackList] = useState(DEFAULT_FIELD_BLACK_LIST);
+
+  useEffect(() => {
+    if (hideField?.length) {
+      const hideValues = hideField.split(',').map(el => el.trim());
+      setFieldBlackList(prev => [...prev, ...hideValues]);
+    }
+  }, [hideField]);
+
+  const searchHandler = event => {
```
v-excelsior, line 28: useCallback

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+    const eventTrigger = event.trigger;
```
v-excelsior, line 29: const eventTrigger = event.trigger;
 const query = event.query;
 
 const {trigger: eventTrigger, query} = event

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+      const query = event.query;
+      const orderedSuggestions = suggestions.filter(suggestion => suggestion.trigger === eventTrigger)[0].suggestions;
+      let finalSuggestions;
+
+      if (!query.trim().length) {
+        finalSuggestions = orderedSuggestions;
+      } else {
+        finalSuggestions = orderedSuggestions
+          .filter(suggestion => suggestion.field.toLowerCase().startsWith(query.toLowerCase()));
+      }
```
v-excelsior, line 45: finalSuggestions = !query.trim()
      ? orderedSuggestions
      : orderedSuggestions.filter(suggestion => compareCaseInsensetive(suggestion, query))

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+    const eventTrigger = event.trigger;
+
+    for (const trigger of triggers) {
+      if (trigger !== eventTrigger) {
+        continue;
+      }
+
+      const query = event.query;
+      const orderedSuggestions = suggestions.filter(suggestion => suggestion.trigger === eventTrigger)[0].suggestions;
+      let finalSuggestions;
```
v-excelsior, line 38: finalSuggestions => filteredSuggestions

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+  const searchHandler = event => {
+    const eventTrigger = event.trigger;
+
+    for (const trigger of triggers) {
+      if (trigger !== eventTrigger) {
+        continue;
+      }
+
+      const query = event.query;
+      const orderedSuggestions = suggestions.filter(suggestion => suggestion.trigger === eventTrigger)[0].suggestions;
```
v-excelsior, line 37: orderedSuggestions => suggestionsByTrigger(?)

Not sure, but it could be a map
```
{
 '@' : <list of suggestions>
 '#' : <list of suggestions>
}
```
What do you think?

components/bl-mention-component/src/index.js
v-excelsior: And all this logic could be moved to function 'filterSuggestions' and here we could keep only one row:

```
setProcessedSuggestions(filterSuggestion(suggestions))
```

I don't spend a lot of time for variable naming, so you could choose one that you like.

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,124 @@
...
+          { fields.map(field => {
+            if (field === 'img' && suggestion.img) {
+              return (<img src={ suggestion[field] } className="img"/>);
+            } else if (suggestion[field]) {
+              return (<span className="text"> { suggestion[field] } </span>);
+            }
+
+            return null;
+          }) }
+        </div>
```
v-excelsior, line 80: Could it be an external component? TriggerCard or something like that(not sure about this name)

components/bl-mention-component/styles/primereact.less
```diff
@@ -0,0 +1 @@
...transform:scale(0)}.p-disabled,.p-disabled *{cursor:default!important;pointer-events:none;user-select:none}.p-component-overlay{position:fixed;top:0;left:0;width:100%;height:100%}.p-connected-overlay{opacity:0;transform:scaleY(.8);transition:transform .12s cubic-bezier(0, 0, .2, 1),opacity .12s cubic-bezier(0, 0, .2, 1)}.p-connected-overlay-visible{opacity:1;transform:scaleY(1)}.p-connected-overlay-hidden{opacity:0;transform:scaleY(1);transition:opacity .1s linear}.p-connected-overlay-enter{opacity:0;transform:scaleY(.8)}.p-connected-overlay-enter-active{opacity:1;transform:scaleY(1);transition:transform .12s cubic-bezier(0, 0, .2, 1),opacity .12s cubic-bezier(0, 0, .2, 1)}.p-connected-overlay-enter-done{transform:none}.p-connected-overlay-exit{opacity:1}.p-connected-overlay-exit-active{opacity:0;transition:opacity .1s linear}.p-inputtext{margin:0}.p-inputtextarea-resizable{overflow:hidden;resize:none}.p-mention{display:inline-flex;position:relative}.p-mention-panel{position:absolute;top:0;left:0;overflow:auto}.p-mention-items{margin:0;padding:0;list-style-type:none}.p-mention-item{cursor:pointer;white-space:nowrap;position:relative;overflow:hidden}.p-fluid .p-mention{display:flex}
```
v-excelsior, line 1: For a case, if it not rendered in the root - please wrap it in our class.
Or if component has api for adding class to this popup - do it.

v-excelsior (changes requested):

components/bl-mention-component/src/index.js
addKiwi: We need to set the order of an image, which should be first.

v-excelsior (changes requested):

components/bl-mention-component/src/suggestion-card.js
```diff
@@ -0,0 +1,13 @@
+export const SuggestionCard = ({ fields, suggestion }) => (
+  <div className="content">
+    { fields.map(field => {
+      if (field === 'img' && suggestion[field]) {
```
Valodya, line 4:
```
if (field === 'img' && suggestion[field]) {
        return (<img src={ suggestion[field] } className="img"/>);
      } 

 if (suggestion[field]) {
        return (<span className="text"> { suggestion[field] } </span>);
      }
```

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,99 @@
...
+const BLACKLISTED_FIELDS = new Set(['created', '___class', 'ownerId', 'updated', 'objectId']);
+
+export default function MentionComponent({ component, eventHandlers }) {
+  const {
+    trigger, suggestions, field, scrollHeight, autoHighlight, placeholder, delay, autoresize, rows, cols,
+    hideField, classList, style, display
+  } = component;
+  const { onChange, onFocus, onBlur, onShow, onHide } = eventHandlers;
+
+  const [suggestionsMap, setSuggestionsMap] = useState(new Map());
```
Valodya, line 19: `useState(() => new Map());`

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,99 @@
...
+import { SuggestionCard } from './suggestion-card.js';
+
+const { Mention } = primereact.mention;
+const { cn } = BackendlessUI.CSSUtils;
+
+const BLACKLISTED_FIELDS = new Set(['created', '___class', 'ownerId', 'updated', 'objectId']);
+
+export default function MentionComponent({ component, eventHandlers }) {
+  const {
+    trigger, suggestions, field, scrollHeight, autoHighlight, placeholder, delay, autoresize, rows, cols,
```
Valodya, line 14: `component.suggestions` means that we have have loaded all suggestions on the client? what if there are +100500 users?

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,99 @@
...
+    if (hasAvailableTrigger) {
+      return (<SuggestionCard fields={ orderedFields } suggestion={ suggestion }/>);
+    }
+
+    setProcessedSuggestions([]);
+
+    return null;
+  }, [blacklistedFields, triggers]);
+
+  const onShowHandler = () => {
```
Valodya, line 66: useCallback

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,99 @@
...
+      style={ style }
+      className={ cn('bl-customComponent-Mention', classList) }
+      scrollHeight={ scrollHeight }
+      autoHighlight={ autoHighlight }
+      placeholder={ placeholder }
+      delay={ delay }
+      rows={ rows }
+      cols={ cols }
+      autoResize={ autoresize }
+      onChange={ e => onChange({ value: e.target.value }) }
```
Valodya, line 92: useCallback

components/bl-mention-component/src/index.js
```diff
@@ -0,0 +1,116 @@
...
+    if (hideField?.length) {
+      const hidenFields = stringToList(hideField);
+      setBlacklistedFields(new Set([...BLACKLISTED_FIELDS, ...hidenFields]));
+    }
+  }, [hideField]);
+
+  const searchHandler = useCallback(event => {
+    const { trigger: eventTrigger, query } = event;
+
+    const outputList = onSearch({ trigger: eventTrigger, query });
```
Valodya, line 39: the `onSearch` event has to return a list of suggestions and inside the event a dev will be able to load it from Backendless or from another storage

PR #193, opened 2023-01-13
MARKET-1591 Accordion

serhmel:


components/bl-accordion-component/src/title.js
```diff
@@ -19,7 +19,9 @@ export function Title({ options }) {
       aria-controls={ `item-${ accordionId }-${ index }-content` }
       role="button"
       tabIndex="0">
-      <i className="material-icons-round" aria-hidden="true">keyboard_arrow_right</i>
+      <svg focusable="false" aria-hidden="true" viewBox="0 0 24 24">
+        <path d="M6.23 20.23 8 22l10-10L8 2 6.23 3.77 14.46 12z"></path>
+      </svg>
```
v-excelsior, line 24: Could you move it as a component to the bottom, and name as <SomethingIcon>? It could make this code part more readable.

Not sure, but if it collapse items - it could be CollapseIcon

components/bl-accordion-component/styles/index.less
```diff
@@ -80,7 +84,7 @@
       }
     }
 
-    &.active i {
+    &.active svg {
```
v-excelsior, line 87: Try to avoid using element selectors if it possible

`.collapse-icon | .collapse-icon-svg` - is more readable, isn't it?

v-excelsior (changes requested):

v-excelsior (approved):

PR #195, opened 2023-01-14
MARKET-1587 Animate on scroll

StasivPavlo:


components/bl-AnimateOnScroll-component/README.md
```diff
@@ -0,0 +1,27 @@
...
+|----------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------|---------------|--------------|------------|-------------------------------------------------------------------------------------------------------------------|
+| Disable        | Select ["mobile", "tablet", "true", "false"]                                                                                                                                                                                                                                                                                                   | "false"       | Disable Logic | YES          | YES        | Allows to determine whether or not to disable the animation                                                       |
```
addKiwi, line 11:
```suggestion
| Disable        | *Select* <br/>[`mobile`, `tablet`, `true`, `false`]                                                                                                                                                                                                                                                                                                   | "false"       | Disable Logic | YES          | YES        | Allows to determine whether or not to disable the animation                                                       |
```
We make select in this way [Example](https://github.com/Backendless/ui-builder-library/blob/bl-AnimateOnScroll-component/components/bl-leafletMap-component/README.md#properties)

addKiwi: Also all types are in italic style ` *Number* `

components/bl-AnimateOnScroll-component/component.json
```diff
@@ -0,0 +1,260 @@
...
+      "hasLogicHandler": true,
+      "handlerId": "anchorLogic",
+      "dataBinding": true,
+      "handlerDescription": "This handler allows determining the element by class whose offset will be used to trigger animation instead of an actual one.",
+      "handlerLabel": "Anchor Logic"
+    },
+    {
+      "type": "number",
+      "name": "offset",
+      "label": "Offset",
```
v-excelsior, line 223: Could we add settingTooltip?

components/bl-AnimateOnScroll-component/component.json
```diff
@@ -0,0 +1,260 @@
...
+      "type": "select",
+      "name": "disable",
+      "label": "Disable",
+      "showInSettings": true,
+      "defaultValue": "false",
+      "hasLogicHandler": true,
+      "handlerId": "disableLogic",
+      "handlerLabel": "Disable Logic",
+      "dataBinding": true,
+      "options": [
```
v-excelsior, line 21: Do we handle case when user returns boolean true from this handler(as it works in all other components) ?

v-excelsior (changes requested):

components/bl-AnimateOnScroll-component/src/index.js
```diff
@@ -0,0 +1,56 @@
...
+
+    document.addEventListener('aos:in', () => {
+      animatedIn();
+    });
+
+    document.addEventListener('aos:out', () => {
+      animatedOut();
+    });
+  }, []);
+
```
v-excelsior, line 34: need to add component.el property with a ref to the root div

components/bl-AnimateOnScroll-component/src/index.js
```diff
@@ -0,0 +1,62 @@
...
+    document.addEventListener('aos:in', () => onAnimationEnter());
+    document.addEventListener('aos:out', () => onAnimationOut());
+
+    return () => {
+      document.removeEventListener('aos:in', () => onAnimationEnter());
+      document.removeEventListener('aos:out', () => onAnimationOut());
+    }
+  }, []);
+
+  component.el = animationOnScrollRef.current;
```
Valodya, line 39: I'm not sure this is the correct way

let's imaging this was rendered only once and the `animationOnScrollRef.current` before first render will be null and as result it won't work 

I assume we need to wrap it with `useEffect`

components/bl-AnimateOnScroll-component/src/index.js
```diff
@@ -0,0 +1,62 @@
...
+  const elementPod = pods['animationPod'];
+
+  const animationOnScrollRef = useRef(null);
+
+  const animationName = useMemo(() => {
+    return animationType + (side === 'none' ? '' : `-${ side }`);
+  }, [animationType, side]);
+
+  useEffect(() => {
+    Aos.init({ disable: DevicesMap[disableFor] });
```
Valodya, line 28: will it work if we modify the `disableFor ` property in runtime?

StasivPavlo: no, that's why I deleted the logic handler and data binding

components/bl-AnimateOnScroll-component/src/index.js
```diff
@@ -0,0 +1,62 @@
...
+
+  useEffect(() => component.el = animationOnScrollRef.current, []);
+
+  if (!display) {
+    return null;
+  }
+
+  return (
+    <div
+      ref={ animationOnScrollRef }
```
Valodya, line 47: `elRef`

components/bl-AnimateOnScroll-component/src/index.js
Valodya: can we specify the `disable` in HTML as `data-aos-disable`?

components/bl-AnimateOnScroll-component/src/index.js
```diff
@@ -0,0 +1,62 @@
...
+const { cn } = BackendlessUI.CSSUtils;
+
+const DevicesMap = {
+  mobile: 'mobile',
+  tablet: 'tablet',
+  all   : true,
+  none  : false,
+};
+
+export default function AnimateOnScroll({ component, eventHandlers, pods }) {
```
Valodya, line 13: have you tried to use more than one instance on the same page with different settings?

components/bl-AnimateOnScroll-component/src/index.js
```diff
@@ -0,0 +1,62 @@
...
+  const animationOnScrollRef = useRef(null);
+
+  const animationName = useMemo(() => {
+    return animationType + (side === 'none' ? '' : `-${ side }`);
+  }, [animationType, side]);
+
+  useEffect(() => {
+    Aos.init({ disable: DevicesMap[disableFor] });
+
+    document.addEventListener('aos:in', () => onAnimationEnter());
```
Valodya, line 30: how does it work when on the page more than one instance?

components/bl-AnimateOnScroll-component/README.md
```diff
@@ -0,0 +1,27 @@
...
+|----------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------|-------------------|--------------|------------|-------------------------------------------------------------------------------------------------------------------|
+| Disable For    | *Select* <br/>[`mobile`, `tablet`, `all`, `none`]                                                                                                                                                                                                                                                                                                     | "false"       | Disable For Logic | YES          | YES        | Allows to determine whether or not to disable the animation                                                       |
```
Valodya, line 11: `false` is the default value for `Disable For`? but in the available list there is only: [`mobile`, `tablet`, `all`, `none`]

components/bl-AnimateOnScroll-component/src/index.js
```diff
@@ -0,0 +1,62 @@
...
+      className={ cn('bl-customComponent-animateOnScroll', classList) }
+      style={ style }
+      data-aos={ animationName }
+      data-aos-easing={ easing }
+      data-aos-duration={ duration }
+      data-aos-delay={ delay }
+      data-aos-offset={ offset }
+      data-aos-mirror={ mirror }
+      data-aos-once={ once }
+      data-aos-anchor={ anchor ? '.' + anchor : null }
```
Valodya, line 57: what does `anchor` look like? is it a className?

components/bl-AnimateOnScroll-component/src/index.js
StasivPavlo: yes, all work well

components/bl-AnimateOnScroll-component/src/index.js
StasivPavlo: no

components/bl-AnimateOnScroll-component/src/index.js
StasivPavlo: AOS dispatches two events on document: aos:in and aos:out whenever any element animates in or out

But I see the possibility of making events for the current element and not for any

components/bl-AnimateOnScroll-component/src/index.js
StasivPavlo: yes

PR #206, opened 2023-01-20
MARKET-1622 Parallax

StasivPavlo:


components/bl-parallax-component/src/index.js
```diff
@@ -20,7 +26,7 @@ export default function Parallax({ component, pods }) {
       <div ref={ backdropRef } className="parallax-background-img"
            style={ { backgroundImage: `url(${ imageUrl })` } }></div>
       <div className="parallax-content">
-        { parallaxContentPod.render() }
+        { parallaxContentPod && parallaxContentPod.render() }
```
v-excelsior, line 29: Could it be like that?

Approach with empty catch looks bad for me.

```suggestion
        { pods['parallaxContent']?.render() }
```

components/bl-parallax-component/src/index.js
```diff
@@ -1,10 +1,9 @@
-import { useRef, useEffect, useCallback } from 'react';
+import { useRef, useEffect, useCallback, useState } from 'react';
```
v-excelsior, line 1: we no longer need useState

v-excelsior (approved):

PR #207, opened 2023-01-20
MARKET-1623 Parallax

StasivPavlo:


components/bl-parallax-component/src/index.js
```diff
@@ -1,15 +1,25 @@
...
   const backdropRef = useRef();
   const containerRef = useRef();
 
-  useAnimation(backdropRef, containerRef, strength);
+  useEffect(() => {
+    if (typeof strength === 'number') {
+      setPower(strength)
+    } else {
+      setPower(0)
+    }
```
v-excelsior, line 19:
```suggestion
      setPower(typeof strength === 'number' ? strength : 0)
```

And if we already use strength - no need in one more definition for it. Use `validStrength` for state variable.

v-excelsior (changes requested):

v-excelsior (approved):

PR #212, opened 2023-01-23
MARKET-1607 Knob

addKiwi:


components/bl-knob-component/src/index.js
```diff
@@ -18,6 +18,9 @@ export default function KnobComponent({ component, eventHandlers }) {
   const [knobDisabled, setKnobDisabled] = useState(!!disabled);
 
   useEffect(() => {
```
v-excelsior, line 20: Are you sure this is working?

For me it's looks not valid.

v-excelsior (changes requested):

components/bl-knob-component/src/index.js
addKiwi: We need it for this case 
image

v-excelsior: Yea, I understand why we need this.

Look please at code - it's not valid, just try to run it.

```
  useEffect(() => {
    setKnobReadOnly(!!readOnly);
  },[readOnly]);

    setKnobDisabled(!!readOnly);
  }, [disabled]);
```

PR #218, opened 2023-01-25
BKNDLSS-31061 Mapbox

addKiwi:


components/bl-mapbox-component/src/helpers.js
```diff
@@ -121,64 +121,77 @@ export const useMarkers = (markers, markersArray, setMarkersArray, mapRef, onMar
...
-          'type': 'geojson',
-          'data': {
-            'type'    : 'Feature',
-            'geometry': {
-              'type'       : 'Polygon',
-              'coordinates': [[...polygon.coordinates]],
-            },
+const addPolygons = (mapRef, polygonsArray, onPolygonClick) => {
+  mapRef.current.on('load', () => {
+    polygonsArray.forEach(polygon => {
```
v-excelsior, line 141: const { description, id, coordinates ... } = polygon

components/bl-mapbox-component/src/helpers.js
```diff
@@ -121,64 +121,77 @@ export const useMarkers = (markers, markersArray, setMarkersArray, mapRef, onMar
...
-          }
-        });
+const updatePolygonsArray = (polygons, mapRef, polygonsArray, setPolygonsArray) => {
+  if (polygons?.length && mapRef.current) {
+    mapRef.current.on('load', () => {
+      polygonsArray.forEach(polygon => {
+        if (mapRef.current.getSource(polygon)) {
+          mapRef.current.removeLayer(`${ polygon.id }-layer`);
+          mapRef.current.removeSource(polygon.id);
+        }
```
v-excelsior, line 131: Can we check if the polygon changed and remove then only if it is?

components/bl-mapbox-component/src/helpers.js
```diff
@@ -121,64 +121,77 @@ export const useMarkers = (markers, markersArray, setMarkersArray, mapRef, onMar
...
-          },
-        });
+      mapRef.current.addLayer({
+        'id'    : `${ polygon.id }-layer`,
+        'type'  : 'fill',
+        'source': polygon.id,
+        'layout': {},
+        'paint' : {
+          'fill-color'  : polygon.color,
+          'fill-opacity': polygon.opacity,
```
v-excelsior, line 160: Can we avoid using notation like 'id' for objects? As for - 'fill-color'. It doesn't work if we use camelCase('fillColor')?

components/bl-mapbox-component/src/helpers.js
```diff
@@ -121,64 +121,77 @@ export const useMarkers = (markers, markersArray, setMarkersArray, mapRef, onMar
...
 
-            const coordinates = { lat: e.lngLat.lat, lng: e.lngLat.lng };
+          const coordinates = { lat: e.lngLat.lat, lng: e.lngLat.lng };
 
-            onPolygonClick({ coordinates: coordinates, description: polygon.description });
-          });
-        } else {
-          mapRef.current.on('click', `${ polygon.id }-layer`, e => {
-            const coordinates = { lat: e.lngLat.lat, lng: e.lngLat.lng };
+          onPolygonClick({ coordinates: coordinates, description: polygon.description });
```
v-excelsior, line 173:
```suggestion
          const {description, ... } = polygon
          
          ...
          
          
          onPolygonClick({ coordinates, description });
```

v-excelsior (changes requested):

components/bl-mapbox-component/src/helpers.js
addKiwi: We are already checking it in usePolygons function.

components/bl-mapbox-component/src/helpers.js
addKiwi: Every polygon needs a unique id to be added on map, I think it is not a good idea to ask users to create a unique name(id) for every polygon. 'fill-color' and 'fill-opacity' doesn't work if we use camelCase.

components/bl-mapbox-component/src/helpers.js
```diff
@@ -1,9 +1,29 @@
...
+const mapOnLoad = (mapRef, listener) => {
+  mapRef.current.on('load', listener);
+};
+
+const mapOnClick = (mapRef, layerId, listener) => {
+  if (layerId) {
+    mapRef.current.on('click', layerId, listener);
+  } else {
+    mapRef.current.on('click', listener);
+  }
```
v-excelsior, line 16: const handlerPayload = layerId ? [layerId, listener] : [listener]

mapRef.current.on('click', ...handlerPayload);

components/bl-mapbox-component/src/helpers.js
```diff
@@ -121,70 +145,92 @@ export const useMarkers = (markers, markersArray, setMarkersArray, mapRef, onMar
...
+      mapAddLayer(mapRef, {
+        'id'    : `${ id }-layer`,
+        'type'  : 'fill',
+        'source': id,
+        'layout': {},
+        'paint' : {
+          'fill-color'  : color,
+          'fill-opacity': opacity,
+        },
       });
```
v-excelsior, line 209: Did you try Class approach for basic Map functions?

components/bl-mapbox-component/src/map.js
```diff
@@ -1,27 +1,26 @@
...
-  const [polygonsArray, setPolygonsArray] = useState([]);
-  const [markersArray, setMarkersArray] = useState([]);
-
   const { accessToken } = settings;
 
   const { onMarkerClick, onPolygonClick } = eventHandlers;
 
   const { markers, polygons, center, classList } = component;
 
+  const map = new Map(mapRef);
```
Valodya, line 18: it creates a new instance on each render and then in the next useEffect it remembers the first one, which means the `initMapboxLibrary ` runs only for the first instance

components/bl-mapbox-component/src/helpers.js
```diff
@@ -51,21 +96,21 @@ export const initMapboxLibrary = (mapRef, mapContainerRef, component, eventHandl
   createActions(mapRef, component);
 
   if (directions) {
-    mapRef.current.addControl(
+    map.addControl(
```
Valodya, line 99: please test changing each property of the component in runtime

PR #226, opened 2023-01-26
MARKET-1634 Markdown

StasivPavlo:


components/bl-markdown-component/src/index.js
```diff
@@ -34,12 +36,17 @@ export default function Markdown({ component }) {
 
   const setContent = text => setMarkdown(md.render(text));
 
+  useEffect(() => {
+    component.el = markdownRef.current;
+  }, []);
```
v-excelsior, line 41: Should we add markdownRef.current to the deps?

v-excelsior (changes requested):

components/bl-markdown-component/src/index.js
Valodya: @StasivPavlo use `elRef` instead

PR #228, opened 2023-01-26
BKNDLSS-31060 Mapbox

addKiwi:


components/bl-mapbox-component/styles/index.less
```diff
@@ -8,4 +8,9 @@
 .bl-customComponent-mapbox {
   height: @bl-customComponent-mapbox-height;
   width: @bl-customComponent-mapbox-width;
+
+  .map-dimensions {
+    width: 100%;
+    height: inherit;
```
addKiwi, line 14: works only with `inherit`.
It fixes the problem with 100% of height, but if the user sets 50% we will see 50% of  map in `div.bl-customComponent-mapbox`

PR #234, opened 2023-01-27
MARKET-1646 Rename some event handlers

StasivPavlo:


components/bl-carousel-component/src/index.js
```diff
@@ -102,8 +102,8 @@ export default function Carousel({ component, eventHandlers }) {
       onMouseLeave={ onMouseLeave }>
       { withControls && (
         <div className="carousel__controls-button">
-          <PrevButton disabled={ !!animation } onPrevButton={ onPrevButton }/>
-          <NextButton disabled={ !!animation } onNextButton={ onNextButton }/>
+          <PrevButton disabled={ !!animation } onPrevButtonClick={ onPrevButtonClick }/>
```
v-excelsior, line 105: Component with name PrevButton mustn't know about onPrevButtonClick action. He should know what to do when onClick triggered.
```suggestion
          <PrevButton disabled={ !!animation } onClick={ onPrevButtonClick }/>
```

components/bl-carousel-component/src/subcomponents.js
```diff
@@ -98,24 +98,24 @@ export function CarouselIndicators(props) {
   );
 }
 
-export function NextButton({ disabled, onNextButton }) {
+export function NextButton({ disabled, onNextButtonClick }) {
```
v-excelsior, line 101: As we already here. Please, refactor that to one component Button which accept type and icon component as props.

v-excelsior (changes requested):

components/bl-carousel-component/src/subcomponents.js
```diff
@@ -98,38 +98,26 @@ export function CarouselIndicators(props) {
   );
 }
 
-export function NextButton({ disabled, onNextButton }) {
+export function Button({ disabled, onClick, type, Icon }) {
```
v-excelsior, line 101: We can avoid passing Icon, as we already pass the type.
Just use the condition to choose the icon inside the component.

Also, I do not clearly understand why we don't use disable property on the button but use disabling pointer events through styles.

StasivPavlo: At first, I was using disable, but later I found a bug that @Valodya helped me solve with pointer events

PR #236, opened 2023-01-30
MARKET-1613 Tree Select returns "empty" text for invalid selected option

serhmel:
Merge only after PR below has been merged.
[MARKET-1615 On Change Event should return value not key](https://github.com/Backendless/ui-builder-library/pull/237)

### DIFF
--- a/components/bl-treeSelect-component/src/helpers.js
+++ b/components/bl-treeSelect-component/src/helpers.js
@@ -95,11 +95,11 @@ export function findNodesInCheckboxMode(node, parent, selectedNodes, expandedKey
 }
 
 export function validateSelectedNodeKeys(selectionMode, selectedNodeKeys, keysMap) {
-  const keys = selectionMode === SelectionMode.single ? { [`${ selectedNodeKeys }`]: true } : { ...selectedNodeKeys };
+  const keys = selectionMode === SelectionMode.single ? { [selectedNodeKeys]: true } : { ...selectedNodeKeys };
 
-  for (const key of Object.keys(keys)) {
+  Object.keys(keys).forEach(key => {
     if (!keysMap[key]) {
       console.error('Error: trying to select non existed option.');
     }
-  }
+  });
 }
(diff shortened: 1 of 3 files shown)


components/bl-treeSelect-component/src/helpers.js
```diff
@@ -93,3 +93,13 @@ export function findNodesInCheckboxMode(node, parent, selectedNodes, expandedKey
     }
   }
 }
+
+export function validateSelectedNodeKeys(selectionMode, selectedNodeKeys, keysMap) {
+  const keys = selectionMode === SelectionMode.single ? { [`${ selectedNodeKeys }`]: true } : { ...selectedNodeKeys };
```
v-excelsior, line 98: ```{ [`${ selectedNodeKeys }`]: true }```
`{ [selectedNodeKeys]: true }`

It's the same, isn't it?

components/bl-treeSelect-component/src/helpers.js
```diff
@@ -93,3 +93,13 @@ export function findNodesInCheckboxMode(node, parent, selectedNodes, expandedKey
...
+
+  for (const key of Object.keys(keys)) {
+    if (!keysMap[key]) {
+      console.error('Error: trying to select non existed option.');
+    }
+  }
```
v-excelsior, line 104:
```
Object.keys(keys).forEach( x => if(..) {
  console.error()
})
```

v-excelsior (changes requested):

components/bl-treeSelect-component/src/helpers.js
```diff
@@ -93,3 +93,13 @@ export function findNodesInCheckboxMode(node, parent, selectedNodes, expandedKey
     }
   }
 }
+
+export function validateSelectedNodeKeys(selectionMode, selectedNodeKeys, keysMap) {
+  const keys = selectionMode === SelectionMode.single ? { [selectedNodeKeys]: true } : { ...selectedNodeKeys };
+
+  Object.keys(keys).forEach(key => {
+    if (!keysMap[key]) {
+      console.error('Error: trying to select non existed option.');
```
v-excelsior, line 102:
```suggestion
      console.error('Can't select a non existed option with key `${<keyValue>}`.');
```

PR #237, opened 2023-01-30
MARKET-1615 On Change Event should return value not key

serhmel:


components/bl-treeSelect-component/src/helpers.js
```diff
@@ -1,105 +1,153 @@
...
+    handleNode(childNode, selectedItems, selectionMode);
+  }
+}
+
+function updateSelectedItems(node, selectedItems, selectionMode) {
+  const { label, data } = node;
+
+  if (selectionMode === SelectionMode.CHECKBOX) {
+    selectedItems[node.key].data = data;
+    selectedItems[node.key].label = label;
```
v-excelsior, line 129:
```suggestion
    selectedItems[node.key] = {... selectedItems[node.key], data, label };
```

But I don't really understand difference between if and else here

components/bl-treeSelect-component/src/index.js
```diff
@@ -54,14 +50,14 @@ export default function TreeSelectComponent({ component, eventHandlers }) {
 
   return (
     <TreeSelect
-      ref={ treeSelectRef }
+      ref={ el => elRef.current = el?.getElement() }
```
v-excelsior, line 53: Do we really need to use assignment here?

As far as I know, `foo = bar` returns `undefined`
So the `ref` prop is also will be undefined?

components/bl-treeSelect-component/src/index.js
```diff
@@ -89,6 +85,20 @@ function useNodeActions(nodes, eventHandlers, setSelectedNodeKey, selectionMode)
...
+    if (!nodes?.length) {
+      return keys;
+    }
+
+    for (const childNode of nodes) {
+      updateKeysMap(childNode, keys);
+    }
+
+    return keys;
+  }, [nodes]);
```
v-excelsior, line 100: Could be moved into hook useKeysMap. For the reason to make component body shorter

components/bl-treeSelect-component/src/index.js
```diff
@@ -119,30 +129,23 @@ function useNodeActions(nodes, eventHandlers, setSelectedNodeKey, selectionMode)
   };
 
   const onSelectedNodeChange = e => {
```
v-excelsior, line 131: Could we add useCallback for some handlers?

v-excelsior (changes requested):

v-excelsior:
We merge two PR into one because of a mistake

components/bl-treeSelect-component/src/helpers.js
serhmel:
```
  if (selectionMode === SelectionMode.CHECKBOX) {
    // selectedItems[node.key] == { checked: true, partialChecked: false }
    selectedItems[node.key] = { ...selectedItems[node.key], data, label };
  } else {
    // selectedItems[node.key] == true
    selectedItems[node.key] = { data, label };
  }
```

components/bl-treeSelect-component/src/index.js
serhmel: if we pass elRef like this

components/bl-treeSelect-component/src/index.js
serhmel: But it is placed not in the component body but in the custom hook.  Do you think it would be better to move it to another hook?

components/bl-treeSelect-component/src/index.js
```diff
@@ -1,33 +1,44 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import primereact from './lib/primereact.min';
+import {
+  ensureMeasure, getSelectedItems, getSelectedKeys, updateKeysMap, updateNode, validateSelectedNodeKeys,
+} from './helpers';
 
 const { cn } = BackendlessUI.CSSUtils;
 const { TreeSelect } = primereact.treeselect;
```
Valodya, line 9: does the lib `primereact ` contain all UI components or only components we use here?

components/bl-treeSelect-component/src/index.js
Valodya: move it to an fn with using useCallback

components/bl-treeSelect-component/src/index.js
serhmel: It contains only the components we used here.

Valodya: ok, great!

PR #265, opened 2023-02-08
MARKET-1685 [Popover Component] Rename handler

StasivPavlo:


components/bl-popover-component/component.json
```diff
@@ -39,15 +39,15 @@
   ],
   "eventHandlers": [
     {
-      "name": "onButtonClick",
-      "label": "On Button Click",
+      "name": "onTriggerClick",
+      "label": "On Trigger Click",
```
v-excelsior, line 43: Trigger -> Target

PR #268, opened 2023-02-08
MARKET-1684 [Popover Component] Add ability to trigger popover by hover over element

StasivPavlo:


components/bl-popover-component/component.json
```diff
@@ -48,6 +48,18 @@
         }
       ],
       "handlerDescription": "This event is triggered when the user clicks on the button."
+    },
+    {
+      "name": "onMouseIn",
+      "label": "On Mouse In",
```
v-excelsior, line 54: image

Not consistent with basic components

PR #273, opened 2023-02-09
MARKET-1669 RegExp doesn't work with mask that contain wild card

StasivPavlo:


components/bl-inputWithMask-component/src/index.js
```diff
@@ -51,7 +51,11 @@ const preparedMask = (maskType, mask) => useMemo(() => {
   }
 
   if (maskType === MaskTypes.REGEX) {
-    return new RegExp(mask);
+    try {
+      return new RegExp(mask);
+    } catch(err) {
+      console.warn(err);
```
v-excelsior, line 57: That's must be an error

v-excelsior (changes requested):

components/bl-inputWithMask-component/src/index.js
v-excelsior, line 57:
```suggestion
      console.error(err);
```

PR #281, opened 2023-02-14
MARKET-1641 Is there a way to make popover position dynamic?

StasivPavlo:


components/bl-popover-component/src/Tooltip.js
```diff
@@ -1,7 +1,8 @@
-import { useEffect, useMemo, useCallback } from 'react';
+import { useEffect, useMemo, useCallback, useState } from 'react';
 import { translatePopover } from './helpers';
 
-export function Tooltip({ contentElement, position, popoverContent }) {
+export function Tooltip({ targetRef, position, popoverContent }) {
+  const [validatePosition, setValidatePosition] = useState(position);
```
v-excelsior, line 5: validate - it's an action, I believe there should be valid or validated

components/bl-popover-component/src/helpers.js
```diff
@@ -1,45 +1,72 @@
...
     [Position.BOTTOM]: {
-      topShift : contentOffset.top + contentHeight - rootOffset.top + root.firstChild.clientHeight,
+      topShift : targetOffset.top + targetHeight - contentOffset.top + contentElement.firstChild.clientHeight,
       leftShift: verticalLeftShift
     }
   };
 
-  return ShiftHandler[position];
+  const PositionCheck = {
+    left  : () => (
```
v-excelsior, line 47: left => Position.LEFT - should be same as for ShiftHandler

components/bl-popover-component/src/helpers.js
```diff
@@ -49,3 +76,21 @@ const getOffset = (el) => {
...
   return { top: rect.top + scrollTop, left: rect.left + scrollLeft };
 };
+
+const validatePosition = (position, PositionCheck, ShiftHandler) => {
+  if (PositionCheck[position]()) {
+    return { ...ShiftHandler[position], newPosition: position };
+  }
+
+  for (const key in Position) {
+    const side = Position[key];
```
v-excelsior, line 86: Object.values(Position).forEach

components/bl-popover-component/src/helpers.js
```diff
@@ -1,45 +1,72 @@
...
     },
     [Position.BOTTOM]: {
-      topShift : contentOffset.top + contentHeight - rootOffset.top + root.firstChild.clientHeight,
+      topShift : targetOffset.top + targetHeight - contentOffset.top + contentElement.firstChild.clientHeight,
       leftShift: verticalLeftShift
     }
   };
 
-  return ShiftHandler[position];
+  const PositionCheck = {
```
v-excelsior, line 46: PositionCheck - PositionValidator

components/bl-popover-component/src/helpers.js
```diff
@@ -1,45 +1,72 @@
...
+        && (targetY + targetHorizontalCenter) >= contentHorizontalCenter
+        && window.innerHeight - targetBottom + targetHorizontalCenter >= contentHorizontalCenter
+    ),
+    top   : () => (
+      targetY >= contentHeight
+        && (targetX + targetVerticalCenter) >= contentVerticalCenter
+        && document.documentElement.clientWidth - targetRight + targetVerticalCenter >= contentVerticalCenter
+    ),
+    bottom: () => (
+      window.innerHeight - targetBottom >= contentHeight
```
v-excelsior, line 63: I'm not sure if it good approach, and my offer is to do it(check for intersection) through IntersectionObserver, but I'm not insisting.

v-excelsior (changes requested):

components/bl-popover-component/src/helpers.js
StasivPavlo: but return doesn't work inside forEach

PR #297, opened 2023-02-20
MARKET-1705 [Knob] Layout is broken when min and max values are negative

addKiwi:


components/bl-knob-component/src/index.js
```diff
@@ -16,6 +16,16 @@ export default function KnobComponent({ component, eventHandlers }) {
...
   const [knobDisabled, setKnobDisabled] = useState(!!disabled);
+  const [knobMaxValue, setKnobMaxValue] = useState(maxValue);
+
+  useEffect(() => {
+    if (maxValue < 0) {
+      setKnobMaxValue(0);
+      console.error('Wrong Max Value');
+    } else {
+      setKnobMaxValue(maxValue);
+    }
```
v-excelsior, line 27:
```suggestion
    if (maxValue < 0) {
      console.error('Maximum value must be greater than zero.');
    }
```

components/bl-knob-component/src/index.js
```diff
@@ -54,7 +64,7 @@ export default function KnobComponent({ component, eventHandlers }) {
         value={ knobValue }
         readOnly={ knobReadOnly }
         disabled={ knobDisabled }
-        max={ maxValue }
+        max={ knobMaxValue }
```
v-excelsior, line 67: max={ Math.max(maxValue. 0) }

No need in additional variable here

v-excelsior (changes requested):

PR #299, opened 2023-02-21
MARKET-1659 [Data Grid] Data grid. Content do not use full space

Saharij:
UI Builder - https://develop.backendless.com/app/zahar_mychka_components/ui-builder/default/pages/data-grid


components/bl-data-grid-component/component.json
```diff
@@ -59,6 +59,18 @@
       "handlerDescription": "This handler allows you to specify whether or not the floating filter is displayed.",
       "handlerLabel": "Floating Filter Logic"
     },
+    {
+      "type": "checkbox",
+      "name": "resizableColumns",
```
v-excelsior, line 64: resizableColumns - name for list that contains only resizable columns. 

Would be better to name this prop - resizable | isColumnsResizable

components/bl-data-grid-component/src/index.js
```diff
@@ -18,6 +19,7 @@ export default function DataGridComponent({ component, eventHandlers }) {
     filter,
     sortable,
     floatingFilter,
+    resizable: resizableColumns,
```
v-excelsior, line 22: At link from PR description nothing change for me, looks like columns still not fill all space.

v-excelsior (changes requested):

components/bl-data-grid-component/src/index.js
Saharij: i forgot to add the modified code to `develop`

PR #304, opened 2023-02-23
BKNDLSS-31536 [Stripe Payment Component] Bugs in the Payment amount field

serhmel:


components/bl-stripe-payment-component/src/payment-components/payment-amount.js
```diff
@@ -1,14 +1,20 @@
...
+export function PaymentAmount({ component, paymentAmount, setPaymentAmount }) {
   const { minAmount, fixedAmount, currency, amountDecimalPlaces } = component;
-  const amountRef = useRef(null);
 
-  const changeAmount = e => {
-    component.amount = validatePaymentAmount(e.target.value, amountDecimalPlaces);
-    amountRef.current.value = amount;
+  const amountPattern = useMemo(() => {
+    return new RegExp('^[0-9]*(\\.[0-9]{0,' + amountDecimalPlaces + '})?$');
+  }, [amountDecimalPlaces]);
```
v-excelsior, line 8: Looks like duplication with components/bl-stripe-payment-component/src/payment-form.js line 18-20
Only dollar sign at end is different

Could it be reusable hook that accept amount and options?

v-excelsior (changes requested):

PR #310, opened 2023-02-24
MARKET-1687 [Radio Buttons List] Add ability to disable individual options

Saharij:
UI Builder - https://develop.backendless.com/app/zahar_mychka_components/ui-builder/default/pages/radio-buttons-list


components/bl-radio-buttons-list-component/src/index.js
```diff
@@ -40,12 +40,13 @@ export default function RadioButtonsListComponent({ component, eventHandlers })
     <div
       style={ style }
```
v-excelsior, line 41: Add elRef?

v-excelsior (changes requested):

PR #319, opened 2023-02-28
MARKET-1706 [Autocomplete] Error when adding grouped options to the component

Saharij:


components/bl-autocomplete-component/src/helpers/use-filtered-options.js
```diff
@@ -1,6 +1,6 @@
 import { useMemo } from 'react';
 
-const compareCaseInsensitive = (str1, str2) => str1.toLowerCase().includes(str2.toLowerCase());
+const compareCaseInsensitive = (str1 = '', str2 = '') => str1.toLowerCase().includes(str2.toLowerCase());
```
v-excelsior, line 3:
```suggestion
const compareCaseInsensitive = (str1, str2) => str1?.toLowerCase().includes(str2.toLowerCase());
```

How about that? Check it please.

v-excelsior (changes requested):

PR #325, opened 2023-03-01
BKNDLSS-31591 [Stripe Payment] It's possible type "+"&"-" sign between numbers into Payment amount field

serhmel:


components/bl-stripe-payment-component/src/payment-components/payment-amount.js
```diff
@@ -1,5 +1,7 @@
 import { useMemo } from 'react';
 
+const plusOrMinusRegex = /[-+]/;
```
v-excelsior, line 3: SIGN_REGEX

PR #333, opened 2023-03-02
BKNDLSS-31607 [Components] Add a mouse pointer effect when the mouse is over the Close button

StasivPavlo:


components/bl-alert-component/styles/index.less
```diff
@@ -106,6 +106,7 @@
 
       background-color: transparent;
       border-radius: 50%;
+      cursor: pointer;
```
v-excelsior, line 109: Need variable for that, also look please if we could add some variable for other controls if they exist

PR #338, opened 2023-03-03
BKNDLSS-31591 [Stripe Payment]It's possible type "+"&"-" sign between numbers into Payment amount field

serhmel:


components/bl-stripe-payment-component/src/payment-components/payment-amount.js
```diff
@@ -1,6 +1,6 @@
 import { useMemo } from 'react';
 
-const ONLY_NUMERIC_REGEX = /[-+eE]/;
+const ONLY_NUMERIC_REGEX = /^[\d.,]+$/;
```
Valodya, line 3: try this pattern `^\d+[.,]?\d+$`

you can test it here https://regex101.com/#javascript

PR #348, opened 2023-03-06
BKNDLSS-31638 [Support] Add actions to Cropper component

serhmel:


components/bl-cropper-component/src/cropper-components.js
```diff
@@ -129,15 +133,11 @@ function Button({ className, label, onClick, visibility, disabled }) {
...
-    return null;
-  }
-
+function ImageUploadButton({ label, onChange, visibility, addImageButtonRef }) {
   return (
-    <label>
+    <label className={ cn({ hidden: !visibility }) }>
       <span>{ label }</span>
-      <input type="file" accept="image/*" onChange={ onChange }/>
+      <input type="file" ref={ addImageButtonRef } accept="image/*" onChange={ onChange }/>
```
v-excelsior, line 140: Could we get rid of this component and move everything that there to HeaderToolbar?

Im think we doesn't loose something if do like this.

components/bl-cropper-component/src/cropper-components.js
```diff
@@ -47,14 +50,15 @@ export function HeaderToolbar({ component, cropperRef, onSave, image, setImage }
...
-  }
-
   return (
-    <div className="header-toolbar">
+    <div className={ cn('header-toolbar', { hidden: !image || !visibility }) }>
       <div className="buttons-wrapper">
-        <ImageUploadButton onChange={ uploadImage } label="Add image" visibility={ addImageButtonVisibility }/>
+        <ImageUploadButton
+          onChange={ uploadImage }
+          label="Add image"
```
v-excelsior, line 58: Also, please add properties for that label, and other if they exist.

v-excelsior (changes requested):

PR #354, opened 2023-03-07
MARKET-1814 Stripe payment component should inform user to add Stripe publishable key

serhmel:


components/bl-stripe-payment-component/src/helpers/use-card-element.js
```diff
@@ -58,7 +69,9 @@ export function useCardElement(eventHandlers, setIsLoading, setTransactionDetail
       .catch(error => {
         const errorMessage = error.message;
 
-        setErrorMessage(errorMessage);
+        setErrorMessage(
+          errorMessage === ErrorMessages.SERVICE_NOT_AVAILABLE ? ErrorMessages.PLUGIN_NOT_AVAILABLE : errorMessage
```
v-excelsior, line 73: Could we make in a little shorter?

ErrorMessages -> Errors

NOT_AVAILABLE -> UNAVAILABLE

v-excelsior (changes requested):

PR #356, opened 2023-03-07
MARKET-1725 [PDF Viewer] Show pagination only when the document is loaded

StasivPavlo:


components/bl-pdfViewer-component/src/index.js
```diff
@@ -49,15 +55,15 @@ export default function PdfViewer({ component, eventHandlers }) {
     }
 
     if (Number(target.value)) {
-      const ensureRange = (v, {min, max}) => Math.max(min, Math.min(v, max));
+      const ensureRange = (v, { min, max }) => Math.max(min, Math.min(v, max));
```
v-excelsior, line 58: Function without dependencies should be declared outside of another function/component

components/bl-pdfViewer-component/src/index.js
```diff
@@ -12,7 +12,7 @@ export default function PdfViewer({ component, eventHandlers }) {
   const [pageIndex, setPageIndex] = useState(1);
   const [documentRef, setDocumentRef] = useState();
   const [pageRef, setPageRef] = useState();
-  const [controlsDisabled, setControlsDisabled] = useState(false);
+  const [isControlDisplay, setIsControlDisplay] = useState(false);
```
v-excelsior, line 15: Maybe will be better to rename it
`isControlVisible` or `displayControl`

Also it's plural, isn't it?

PR #361, opened 2023-03-08
MARKET-1819 [Data Grid] Dimension fields can't take on values without "px".

Saharij:
UI Builder - https://develop.backendless.com/app/zahar_mychka_components/ui-builder/default/pages/data-grid

### DIFF
--- a/components/bl-data-grid-component/src/use-styles.js
+++ b/components/bl-data-grid-component/src/use-styles.js
@@ -21,6 +21,8 @@ const validateStyles = dimensions => {
 };
 
 const validateDimensions = dimensions => {
+  const numbersOnlyRegex = /^\d+$/;
+
   return Object.entries(dimensions).reduce((acc, [propertyName, value]) => {
     if (!value) {
       console.error(`Property ${ propertyName } has wrong value!`);
@@ -28,10 +30,10 @@ const validateDimensions = dimensions => {
       return acc;
     }
 
-    if (value.includes("%")) {
-      return { ...acc, [propertyName]: `${ parseFloat(value) }%` };
+    if (numbersOnlyRegex.test(value)) {
+      return { ...acc, [propertyName]: `${ value }px` };
     }
 
-    return { ...acc, [propertyName]: `${ parseFloat(value) }px` };
+    return { ...acc, [propertyName]: value };
   }, {});
 };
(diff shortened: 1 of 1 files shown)


components/bl-data-grid-component/src/use-styles.js
```diff
@@ -6,17 +6,32 @@ export const useStyles = (style, width, height) => useMemo(() => {
...
+      return acc;
+    }
+
+    if (value.includes("%")) {
+      return { ...acc, [propertyName]: `${ parseFloat(value) }%` };
+    }
+
+    return { ...acc, [propertyName]: `${ parseFloat(value) }px` };
+  }, {});
+};
```
v-excelsior, line 37: Looks like redundant logic. Could you describe cases when it's needed?

I think `CSS.supports(propertyName, value)` it's all what we need.

components/bl-data-grid-component/src/use-styles.js
```diff
@@ -6,17 +6,34 @@ export const useStyles = (style, width, height) => useMemo(() => {
...
+      return acc;
+    }
+
+    if (numbersOnlyRegex.test(value)) {
+      return { ...acc, [propertyName]: `${ value }px` };
+    }
+
+    return { ...acc, [propertyName]: value };
+  }, {});
+};
```
v-excelsior, line 39: There should be only one iteration over the object.

below is pseudocode

```
for each property in dimensions {
   check for any not numeric character in the end
   
   if false { 
     ensure px with warning about our autocomplete( <property name> automatically changed: 50 -> 50px as it's invalid CSS value.)
   }
   
   check CSS suports
   
  return valid property or nothing
}

```

components/bl-data-grid-component/src/use-styles.js
```diff
@@ -6,16 +6,21 @@ export const useStyles = (style, width, height) => useMemo(() => {
   return { ...style, ...validStyles, flexShrink: 0 };
 }, [style, width, height]);
 
-const validateStyles = styles => {
-  return Object.entries(styles).reduce((acc, [propertyName, value]) => {
+const validateStyles = dimensions => {
+  const numbersOnlyRegex = /^\d+$/;
```
v-excelsior, line 10:
```suggestion
  const ONLY_NUMERIC_REGEX = /^\d+$/;
```
And move out from function

PR #364, opened 2023-03-09
MARKET-1726 [PDF Viewer] Width and height should apply to the entire component

StasivPavlo:


components/bl-pdfViewer-component/src/index.js
```diff
@@ -47,8 +50,10 @@ export default function PdfViewer({ component, eventHandlers }) {
   };
 
   const onPageLoadSuccess = () => {
+    const spaceForControls = getBottomOffset(controlsRef.current) - getBottomOffset(pageRef);
+
+    pageRef.firstChild.style.height = `calc(${ height } - ${ spaceForControls }px)`;
```
v-excelsior, line 55: Why do we need calc here?

StasivPavlo: need to specify the dimensions for the element that displays the pdf, if we use height that comes from the property, the block with control elements will fall out of the root element. Therefore, from the height that comes from the properties, I subtract the necessary space for displaying the control elements and assign it to the element that displays the pdf

StasivPavlo: I subtracted it through calc because `height` comes with a unit of measurement (px, %, etc.), and `spaceForControls` is just a number

PR #366, opened 2023-03-10
MARKET-1605 Barcode component

serhmel:
[UI Builder]( https://develop.backendless.com/app/serhiy_melnychuk_components/ui-builder/Barcode/pages/landing)
[Desktop Preview]( https://genteelinstrument.backendless.app/api/files/ui-builder/containers/Barcode/index.html?page=landing)

page_preview

UI Builder Preview:
ui_biuilder_preview


Saharij (commented):
LGTM🔥

components/bl-barcode-component/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+const FontOptions = {
+  UNSET      : 'unset',
+  BOLD       : 'bold',
+  ITALIC     : 'italic',
+  BOLD_ITALIC: 'boldItalic',
+};
+
+const FontOptionsMap = {
+  [FontOptions.UNSET]      : '',
```
v-excelsior, line 15: I'm not sure, but we have an issue with empty string at select trait's in default components. To check if it not reproduced try to: Reload Page change value from 'bold' to '' in settings and check in preview if it changed and visa verse(reload, change from '' to 'bold' and check again)

components/bl-barcode-component/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+export default function BarcodeComponent({ component, eventHandlers, elRef }) {
+  const { classList, display, style } = component;
+  const { onClick, onMouseOver, onMouseOut } = eventHandlers;
+
+  const barcodeRef = useRef(null);
+
+  useBarcodeLibrary(component, barcodeRef);
+
+  const styles = {
+    display: !display && 'none',
```
v-excelsior, line 30: Why we use `display:none` but not `return null`?

components/bl-barcode-component/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+  const onValidate = useCallback(valid => {
+    if (!valid) {
+      console.error(`"${ value }" is not a valid value for the ${ format } barcode format.`);
+    }
+  }, [value, format]);
+
+  useEffect(() => {
+    if (value) {
+      JsBarcode(barcodeRef.current, value, { ...options, valid: onValidate });
```
v-excelsior, line 74: What if the value is the same but the options are changed? It looks like this useEffect will not be running again.

v-excelsior (changes requested):

components/bl-barcode-component/src/index.js
serhmel: Everything works fine.
We don't use an empty string at ui builder select, we use an 'unset' value for it. And when it's selected, we pass an empty string in the library's font parameters.

components/bl-barcode-component/src/index.js
serhmel: Due to the fact that we pass the ref to the library to generate a barcode, and if it does not render, the page load will break due to a library issue.

To avoid this we can add a condition before calling barcode generator and add "display" property to dependencies array.
```
  useEffect(() => {
    if (!barcodeRef.current) {
      return;
    }

    if (value) {
      JsBarcode(barcodeRef.current, value, { ...options, valid: onValidate });
    } else {
      barcodeRef.current.removeAttribute('src');
    }
  }, [display, value]);
```

PR #369, opened 2023-03-10
MARKET-1734 [Double Calendar] Default opening for double calendar

Saharij:

### DIFF
--- a/components/bl-double-calendar-component/src/date-range.js
+++ b/components/bl-double-calendar-component/src/date-range.js
@@ -2,6 +2,14 @@ import { useState, useEffect } from 'react';
 
 import DatePicker from './lib/react-datepicker.min.js';
 
+const { cn } = BackendlessUI.CSSUtils;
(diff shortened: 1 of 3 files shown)


components/bl-double-calendar-component/src/date-range.js
```diff
@@ -2,6 +2,14 @@ import { useState, useEffect } from 'react';
 
 import DatePicker from './lib/react-datepicker.min.js';
 
+const { cn } = BackendlessUI.CSSUtils;
+
+// this is needed to display the current month in the calendar after the reset without selecting the current day
+const setCurrentMonth = setState => {
+  setState(new Date());
+  setTimeout(() => setState(null), 1);
+};
```
v-excelsior, line 11: I don't think it is a good practice when we create a setter function which will accept another setter. Than not looks obvious.  Will be better to create handlers using useCallback inside the component.

PR #386, opened 2023-03-15
MARKET-1829 [PDF Viewer Component] It's impossible to enter a page number if the number of sheets in the file is less than 10

StasivPavlo:


components/bl-pdfViewer-component/src/subcomponents.js
```diff
@@ -21,7 +23,31 @@ export function NoData() {
...
+    if (Number(inputValue)) {
+      const page = ensureRange(inputValue, { min: 1, max: numPages });
+
+      setPageIndex(page);
+      setInputValue(page);
+    } else {
+      setPageIndex(1);
+      setInputValue(1);
+    }
+  };
```
v-excelsior, line 40: add useCallback

rename handlerSubmit -> onSubmit

components/bl-pdfViewer-component/src/subcomponents.js
```diff
@@ -21,7 +23,31 @@ export function NoData() {
...
+
+  const handleKeyDown = event => {
+    if (event.key === 'Enter') {
+      handlerSubmit();
+    }
+  };
```
v-excelsior, line 46: handleKeyDown -> onKeyDown | onEnter

components/bl-pdfViewer-component/src/subcomponents.js
```diff
@@ -21,7 +23,31 @@ export function NoData() {
 }
 
 export function Controls(props) {
-  const { pageIndex, setPageIndex, inputRef, handlerPageChange, numPages, display, controlsRef } = props;
+  const { pageIndex, setPageIndex, inputRef, numPages, display, controlsRef } = props;
+
+  const [inputValue, setInputValue] = useState(1);
```
v-excelsior, line 28: inputValue -> nextPage

components/bl-pdfViewer-component/src/subcomponents.js
```diff
@@ -21,7 +23,31 @@ export function NoData() {
...
+
+  const handlerSubmit = () => {
+    if (Number(inputValue)) {
+      const page = ensureRange(inputValue, { min: 1, max: numPages });
```
v-excelsior, line 32: numPages is page count? If yes, rename it please for pageCount

components/bl-pdfViewer-component/src/subcomponents.js
```diff
@@ -21,7 +23,31 @@ export function NoData() {
...
+      setInputValue(page);
+    } else {
+      setPageIndex(1);
+      setInputValue(1);
+    }
+  };
+
+  const handleKeyDown = event => {
+    if (event.key === 'Enter') {
+      handlerSubmit();
```
v-excelsior, line 44: could we get `event.target.value` here and pass into submit handler?

components/bl-pdfViewer-component/src/subcomponents.js
```diff
@@ -39,8 +65,10 @@ export function Controls(props) {
         <input
           ref={ inputRef }
           className="page-input"
-          value={ pageIndex }
-          onChange={ handlerPageChange }
+          value={ inputValue }
+          onChange={ ({ target }) => setInputValue(target.value) }
```
v-excelsior, line 69: I believe we could avoid using addational state for inputValue here and make this input uncotroller. Could you try to do it?

v-excelsior (changes requested):

components/bl-pdfViewer-component/src/subcomponents.js
StasivPavlo: I'm not sure what is needed use useCallback on native HTML elements.
https://github.com/facebook/react/issues/17055

components/bl-pdfViewer-component/src/index.js
```diff
@@ -9,7 +9,7 @@ export default function PdfViewer({ component, eventHandlers, elRef }) {
   const { style, display, classList, pdfUrl, width, height } = component;
   const { onLoadSuccess, onLoadError } = eventHandlers;
 
-  const [numPages, setNumPages] = useState(null);
+  const [pageCount, setPageCount] = useState(null);
   const [pageIndex, setPageIndex] = useState(1);
```
v-excelsior, line 13: pageIndex -> currentPage ?

components/bl-pdfViewer-component/src/subcomponents.js
```diff
@@ -32,28 +58,31 @@ export function Controls(props) {
...
+        disabled={ (pageIndex > pageCount) || (pageIndex <= 1) }>
         <PrevButtonIcon/>
       </button>
       <div className="pages-info">
         <input
           ref={ inputRef }
           className="page-input"
-          value={ pageIndex }
-          onChange={ handlerPageChange }
+          defaultValue={ 1 }
```
v-excelsior, line 68: Looks like we doesn't make it better, and it's still controlled but via the ref. Let revert to using state and controlled input, it's more obvious. Sorry for inconveniences.

components/bl-pdfViewer-component/src/subcomponents.js
```diff
@@ -21,7 +23,31 @@ export function NoData() {
...
+  const onSubmit = ({ target }) => {
+    if (Number(target.value)) {
+      const page = ensureRange(target.value, { min: 1, max: pageCount });
+
+      setPageIndex(page);
+      target.value = page;
+    } else {
+      setPageIndex(1);
+      target.value = 1;
+    }
```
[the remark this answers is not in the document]
v-excelsior: Also, I not sure, but why we reset it to 1? Maybe will be better to stay at the same page what will be earlier?

For example, we at 5 page, document has 10 pages. And we try to set page 99. Now it will go to 1 page, but better to stay at 5, and make warning/error in console. Or even better, highlith the input and make tooltip that says about non existed page and real range of pages.

`Non existed page "{pageThatWeTryToApply}". Please, choose the page in range {minPage} - {maxPage}`

v-excelsior (changes requested):

components/bl-pdfViewer-component/src/subcomponents.js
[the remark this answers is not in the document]
StasivPavlo: we reset to 1 page only if the value entered is less than 1, or if it is not a number but a string. If we enter a value greater than possible, the last page will open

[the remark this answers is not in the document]
StasivPavlo: and I think it is better to further limit the input of non-numeric values

PR #387, opened 2023-03-15
BKNDLSS-31724 [MapBox Component] The polygon is not displayed after updating the data for it

addKiwi:


components/bl-mapbox-component/src/helpers.js
```diff
@@ -219,8 +219,8 @@ const createPopup = (polygon, mapRef, onPolygonClick, map) => {
   }
 };
 
-const addPolygons = (mapRef, polygonsArray, onPolygonClick, map) => {
-  map.onLoad(() => {
+const addPolygons = (mapRef, polygonsArray, onPolygonClick, map, ready) => {
+  if (ready) {
```
v-excelsior, line 223: Could we not pass `ready` here, and use it earlier?

```
  useEffect(() => {
    if(ready) {
        addPolygons(mapRef, polygonsArray, onPolygonClick, map);
    }
  }, [polygonsArray, ready]);

Comment (v-excelsior) [reply components/bl-mapbox-component/src/helpers.js:223]: Also same for `updatePolygonsArray`

File: components/bl-mapbox-component/src/map.js [hunk 2]
```diff
@@ -8,6 +8,7 @@ const { cn } = BackendlessUI.CSSUtils;
 export default function MapboxComponent({ component, eventHandlers, settings }) {
   const mapContainerRef = useRef(null);
   const mapRef = useRef(null);
+  const [ready, setReady] = useState(false);
```
v-excelsior, line 11: ready -> `isMapLoaded`

v-excelsior (changes requested):

PR #395, opened 2023-03-17
MARKET-1726 [PDF Viewer] Width and height should apply to the entire component

StasivPavlo:


components/bl-pdfViewer-component/src/index.js
```diff
@@ -27,17 +30,24 @@ export default function PdfViewer({ component, eventHandlers, elRef }) {
...
+    }
+  }, [pageLoaded, elRef, height, width]);
+
   useEffect(() => {
     if (pageLoaded) {
       const spaceForControls = getBottomOffset(controlsRef.current) - getBottomOffset(pageRef);
 
-      documentRef.style.height = `calc(${ height } - ${ spaceForControls }px)`;
-      documentRef.style.width = width || 'auto';
+      documentRef.style.height = rootHeight - spaceForControls + 'px';
```
v-excelsior, line 44: We could make it more obvious

`${rootHeight - spaceForControls}px`

same below

v-excelsior: Also, it could be an utility function like

`usePx` `addPx` or simply `px(variable)`, measure(variable, 'px') choose which you like

v-excelsior (changes requested):

PR #396, opened 2023-03-17
MARKET-1733 [Double Calendar] Add date format selection in settings

Saharij:


components/bl-double-calendar-component/src/helpers/utils.js
```diff
@@ -13,3 +13,21 @@ export function differenceInDays(start, end) {
...
+    return null;
+  }
+
+  const splitDate = date.split("-");
+  const { DD, MM, YYYY } = format.split("-").reduce((acc, item, index) => {
+    return {
+      ...acc,
+      [item]: index,
+    };
+  }, {});
```
v-excelsior, line 30: getFormatPartsOrder

components/bl-double-calendar-component/src/helpers/utils.js
```diff
@@ -13,3 +13,21 @@ export function differenceInDays(start, end) {
...
 }
+
+export function normalizeFormat(date, format) {
+  if (!date) {
+    console.warn('Date is not provided!')
+
+    return null;
+  }
+
+  const splitDate = date.split("-");
```
v-excelsior, line 24: splitDate -> dateParts

components/bl-double-calendar-component/src/index.js
```diff
@@ -29,8 +31,8 @@ export default function DoubleCalendarComponent({ component, eventHandlers }) {
         />
         :
         <DateRange
-          toDate={ toDate }
-          fromDate={ fromDate }
+          toDate={ normalizeFormat(toDate, dateFormat) }
```
v-excelsior, line 34: ```const formatNormalizer = useMemo(()=> date => normalizeFormat(dateFormat, date))```

Change args order and use closures here. Also both ranges could be memoized too.
Contact me if something unclear.

components/bl-double-calendar-component/src/helpers/use-actions.js
```diff
@@ -1,18 +1,23 @@
...
+      const normalizedFromDate = new Date(normalizeFormat(fromDate, dateFormat));
+      const normalizedToDate = new Date(normalizeFormat(toDate, dateFormat));
+
+      const diffInTime = differenceInTime(normalizedFromDate, normalizedToDate);
 
       if (diffInTime >= 0) {
-        setStartDate(new Date(fromDate));
-        setEndDate(new Date(toDate));
+        setStartDate(normalizedFromDate);
+        setEndDate(normalizedToDate);
```
v-excelsior, line 20: I assume `normalizedFromDate` and `normalizedToDate` could be passed as argument here, check it please.

components/bl-double-calendar-component/src/helpers/use-actions.js
```diff
@@ -1,18 +1,23 @@
...
       } else {
         console.warn("Date is not valid!");
```
v-excelsior, line 22: Please add reason here.

v-excelsior (changes requested):

components/bl-double-calendar-component/src/helpers/use-actions.js
Saharij: Dates that passed to this function came from user input, that's why we need to normalize it here. If I understood you correctly.

PR #400, opened 2023-03-20
MARKET-1838 [Double Calendar]Display selected dates in header

Saharij:


components/bl-double-calendar-component/src/helpers/utils.js
```diff
@@ -13,3 +13,13 @@ export function differenceInDays(start, end) {
...
 }
+
+export function getShortDate(date) {
+  const dateObject = new Date(date);
+
+  const month = dateObject.toLocaleString('en-US', { month: 'short' });
+  const day = dateObject.getDate();
+  const year = dateObject.getFullYear();
+
+  return `${month} ${day} ${year}`;
```
v-excelsior, line 24: date.toLocaleDateString()

That will helps us to avoid wrong format depend on locale

v-excelsior (changes requested):

components/bl-double-calendar-component/src/helpers/utils.js
Saharij: If do not pass the locale, it will turn out like this
Screenshot 2023-03-20 at 19 04 16

Saharij: I think it will be better if `month` is `short` because
Screenshot 2023-03-21 at 11 07 10

PR #405, opened 2023-03-22
[PDF Viewer] Error after changing the display to false

StasivPavlo:
image

### DIFF
--- a/components/bl-pdfViewer-component/src/index.js
+++ b/components/bl-pdfViewer-component/src/index.js
@@ -25,8 +25,8 @@ export default function PdfViewer({ component, eventHandlers, elRef }) {
 
   useEffect(() => {
     if (documentRef) {
-      documentRef.style.height = height;
-      documentRef.style.width = width;
+      documentRef.style.height = '100%';
+      documentRef.style.width = '100%';
     }
   }, [documentRef, height, width]);
 
@@ -35,7 +35,7 @@ export default function PdfViewer({ component, eventHandlers, elRef }) {
       setRootHeight(elRef.current.getBoundingClientRect().height);
       setRootWidth(elRef.current.getBoundingClientRect().width);
     }
-  }, [pageLoaded, elRef, height, width]);
+  }, [pageLoaded, height, width]);
 
   useEffect(() => {
     if (pageLoaded) {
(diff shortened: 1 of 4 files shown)


components/bl-pdfViewer-component/src/index.js
```diff
@@ -25,8 +25,8 @@ export default function PdfViewer({ component, eventHandlers, elRef }) {
 
   useEffect(() => {
     if (documentRef) {
-      documentRef.style.height = height;
-      documentRef.style.width = width;
+      documentRef.style.height = '100%';
+      documentRef.style.width = '100%';
     }
   }, [documentRef, height, width]);
```
v-excelsior, line 31: We still need height and width in deps?

components/bl-pdfViewer-component/src/index.js
```diff
@@ -35,7 +35,7 @@ export default function PdfViewer({ component, eventHandlers, elRef }) {
       setRootHeight(elRef.current.getBoundingClientRect().height);
       setRootWidth(elRef.current.getBoundingClientRect().width);
     }
-  }, [pageLoaded, elRef, height, width]);
+  }, [pageLoaded, height, width]);
```
v-excelsior, line 38: same here

> We still need height and width in deps?

components/bl-pdfViewer-component/src/index.js
```diff
@@ -123,9 +130,13 @@ export default function PdfViewer({ component, eventHandlers, elRef }) {
...
 const getBottomOffset = el => {
-  const rect = el.getBoundingClientRect();
+  if (el) {
+    const rect = el.getBoundingClientRect();
+
+    return rect.bottom + window.scrollY;
+  }
 
-  return rect.bottom + window.scrollY;
+  return null;
```
v-excelsior, line 139: why null?

For example here:

`const spaceForControls = getBottomOffset(controlsRef.current) - getBottomOffset(pageRef);`

could be `null - 10` - looks strage

v-excelsior (changes requested):

components/bl-pdfViewer-component/styles/index.less
```diff
@@ -111,6 +111,10 @@
     background-color: #aaa;
   }
 
+  .no-data-text {
+    text-align: center;
```
v-excelsior, line 115: Also will be good to add some variables. 

color, font-size, text-align

And change this element from h3 to span. Or explain why that should be heading. Using it for styling purposes it's not a good idea.

PR #416, opened 2023-03-28
MARKET-1770 Calendar Heat Map. Calendar data destroing view

StasivPavlo:


components/bl-calendarHeatmap-component/src/helpers.js
```diff
@@ -1,9 +1,33 @@
...
+      const previousDay = getPreviousDay(today, i);
+      const foundData = data.find(value => new Date(value.date).toDateString() === previousDay.toDateString());
+
+      if (foundData) {
+        const { date, count } = foundData;
+
+        newData.push({ date, count });
+      } else {
+        newData.push({ date: previousDay, count: 0 });
+      }
```
v-excelsior, line 17:
```
const { date, count } = foundData || { date: previousDay, count: 0 }

newData.push({ date, count });
```

components/bl-calendarHeatmap-component/src/helpers.js
```diff
@@ -1,9 +1,33 @@
-export const validate = items => {
-  if (typeof items === 'string') {
-    return items.split(',');
+import { today } from './index';
+
+export const generateData = (numberDays, data) => {
+  if (numberDays && data) {
```
v-excelsior, line 4: move this check higher(before function calling)

components/bl-calendarHeatmap-component/src/helpers.js
```diff
@@ -1,9 +1,33 @@
...
+    const newData = [];
+
+    for (let i = 0; i < numberDays; i++) {
+      const previousDay = getPreviousDay(today, i);
+      const foundData = data.find(value => new Date(value.date).toDateString() === previousDay.toDateString());
```
v-excelsior, line 9:
```suggestion
      const foundData = data.find(({date}) => new Date(date).toDateString() === previousDay.toDateString());
```

Also, why we compare strings? How about `valueOf`?
Also it could be utility function - `isSameDate` for example

v-excelsior (changes requested):

components/bl-calendarHeatmap-component/src/helpers.js
StasivPavlo: _Also, why we compare strings? How about valueOf?_
valueOf is not suitable because valueOf includes time

PR #425, opened 2023-03-30
BKNDLSS-31858 [Multiple Select] The "Select all" checkmark is scrolled up with other list.

Saharij:


components/bl-multiple-select-component/styles/index.less
```diff
@@ -70,13 +70,23 @@
...
     border-radius: 4px;
+    margin-top: 60px;
+    inset: 0 auto auto 0;
     background-color: @bl-customComponent-multipleSelect-backgroundColor;
     color: @bl-customComponent-multipleSelect-textColor;
-    overflow: auto;
     z-index: 100;
+
+    &__list {
+      max-height: 200px;
```
v-excelsior, line 82: Could we add variable for that?

components/bl-multiple-select-component/styles/index.less
```diff
@@ -70,13 +70,23 @@
   .options {
     position: absolute;
     width: 300px;
```
v-excelsior, line 72: Could we add variable for that?

components/bl-multiple-select-component/styles/index.less
```diff
@@ -243,9 +253,11 @@
     background-color: @bl-customComponent-selectAllCheckbox-backgroundColor;
     padding: 6px 16px;
     box-sizing: border-box;
-    border-bottom: 2px solid @bl-customComponent-multipleSelect-themeColor;
     overflow: hidden;
     cursor: pointer;
+    border-bottom: 2px solid @bl-customComponent-multipleSelect-themeColor;
+    border-top-left-radius: 4px;
+    border-top-right-radius: 4px;
```
v-excelsior, line 260: Also variable

v-excelsior (changes requested):

components/bl-multiple-select-component/src/options/index.js
```diff
@@ -1,21 +1,23 @@
...
 
+const { cn } = BackendlessUI.CSSUtils;
+
 const DEFAULT = 'default';
 
 export function Options(props) {
   const { type, options, selectValue, selectAllCheckbox, selectAllLabel, setSelectValue, onChange } = props;
   const selectRef = useRef(null);
-  const [margin, setMargin] = useState(0);
+  const [optionsPlacement, setOptionsPlacement] = useState('bottom');
```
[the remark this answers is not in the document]
Saharij: And what if change the state logic? Change `OptionsPlacement` to an `isOptionsOnTop`. In this case, will not need an `OptionPlacement` map. Because the main task of this state is to check if the options are on top.

Does this make sense?

PR #431, opened 2023-03-31
BKNDLSS-31857 [Multiple Select] Console error on the example created option.

Saharij:


components/bl-multiple-select-component/src/helpers/validate.js
```diff
@@ -1,6 +1,6 @@
 export const validateOptions = options => {
   if (!Array.isArray(options)) {
-    console.error('Options should be a list');
+    console.error('Options should be a list OR you have an empty “Options Logic” handler in Logic, delete or fill it.');
```
v-excelsior, line 3: That looks to me like we are trying to prevent basic things. When the dev sees the error that says - something is wrong with your options. There are only two places where it could come: settings and logic. 

I'm not saying we are making some adverse thing, but it looks strange if we add so "low level" error only here. 

As for me, would be better something like this:
`Invalid options type. Expected list but got ${current type}`

P.S.: if we choose approach that we see in PR, I would rather see something like
`Options should be a list. Check the Options Logic/setting.`

PR #432, opened 2023-03-31
BKNDLSS-31864 [Multiple Select] Layering in couple places.

Saharij:


components/bl-multiple-select-component/src/options/option.js
```diff
@@ -10,7 +10,7 @@ export function Option({ type, option, isOptionSelected, handleSelectValue }) {
       onClick={ () => handleSelectValue(option) }
       className={ cn("option", { "option__selected": isOptionSelected }) }>
       { type === CHECKMARK && <Checkmark checked={ isOptionSelected } /> }
-      <span>{ option.label }</span>
+      <span className="option__label-truncate">{ option.label }</span>
```
v-excelsior, line 13: Just for clarifying. Do you use some kind of BEM?

If yes, that should be:

`block__element_modification` or `block_element__modification` or `block--element_modification` and so on.
As you see the main idea is split block - element AND element - modification using different separators. But here I can see we use same separator(`__`) for both.

As for  `option__label-truncate`. Do we have another type of label? If no - there are no need to name this class using modificator(`label-truncate`). That should be just `label` or we could implement utility class `text-truncate` that could be applied to any span.

v-excelsior (changes requested):

PR #456, opened 2023-04-06
BKNDLSS-32009 [Audio Player] Bad UI for the stream

serhmel:


components/bl-audio-player-component/src/index.js
```diff
@@ -58,11 +54,7 @@ export default function AudioPlayer({ component }) {
   component.playAudio = () => setIsPlaying(true);
   component.stopAudio = () => stopPlaying();
   component.replaceAudio = (audioUrl, audioTitle) => {
-    setCurrentTrack({
-      url  : audioUrl,
-      title: audioTitle,
-    });
-
+    setCurrentTrack({ url: audioUrl, title: audioTitle });
```
v-excelsior, line 57: You could rename function argument to achieve

`setCurrentTrack({ url, title });`

components/bl-audio-player-component/src/player-components/track-time.js
```diff
@@ -16,8 +16,12 @@ export function TrackTime({ audioRef, timer }) {
...
     <div className="track-time">
       <div>{ timer }</div>
-      <span>/</span>
-      <div>{ trackLength }</div>
+      { trackLength && (
+        <span>/</span>
+      ) }
+      { trackLength && (
+        <div>{ trackLength }</div>
+      ) }
```
v-excelsior, line 24:
```
      { trackLength && (
      <>
        <span>/</span>
        <div>{ trackLength }</div>
      </>
      ) }
```

or 

```
      { trackLength && (<span>/ { trackLength }</span>) }
```
      
I prefer second if there are no need in div

serhmel: I chose the first variant because the space between the elements is controlled by the span margin.

PR #467, opened 2023-04-12
MARKET-1893 [Popover] Add delay for the On Mouse Over Event & On Mouse Out Event.

StasivPavlo:


components/bl-popover-component/src/index.js
```diff
@@ -15,6 +16,14 @@ export default function Popover({ component, eventHandlers, pods }) {
 
   component.setIsOpen = setIsOpen;
 
+  const onMouseEnter = () => {
+    setTimeout(() => onMouseOver({ isOpen }), delayMouseOver);
+  };
+
+  const onMouseLeave = () => {
+    setTimeout(() => onMouseOut({ isOpen }), delayMouseOut);
+  };
```
v-excelsior, line 25: I assume there will be a bug when we make delay for example 5s and spam enter and leave.
We should create and remember timeout, and if new was created but old not yet runned, we should remove old.

components/bl-popover-component/src/helpers.js
```diff
@@ -46,30 +46,30 @@ export const translatePopover = (targetRef, contentElement, position) => {
...
+      && (targetY + targetHorizontalCenter) >= contentHorizontalCenter
+      && window.innerHeight - targetBottom + targetHorizontalCenter >= contentHorizontalCenter
     ),
     [Position.TOP]   : () => (
       targetY >= contentHeight
-        && (targetX + targetVerticalCenter) >= contentVerticalCenter
-        && document.documentElement.clientWidth - targetRight + targetVerticalCenter >= contentVerticalCenter
+      && (targetX + targetVerticalCenter) >= contentVerticalCenter
+      && document.documentElement.clientWidth - targetRight + targetVerticalCenter >= contentVerticalCenter
     ),
```
v-excelsior, line 61: Could we rewrite it as below?

```
    [Position.TOP]   : () => {
      return targetY >= contentHeight
        && (targetX + targetVerticalCenter) >= contentVerticalCenter
        && document.documentElement.clientWidth - targetRight + targetVerticalCenter >= contentVerticalCenter
    },
    ```

File: components/bl-popover-component/src/helpers.js [hunk 3]
```diff
@@ -19,7 +19,7 @@ export const translatePopover = (targetRef, contentElement, position) => {
 
   const {
     width: targetWidth, height: targetHeight,
-    x: targetX, y: targetY,
+    x    : targetX, y: targetY,
```
v-excelsior, line 22: looks strange, could we return as it was?

v-excelsior (changes requested):

PR #469, opened 2023-04-13
MARKET-1693 [Modal] Visibility logic does not work

Saharij:
[UI Builder](https://develop.backendless.com/app/zahar_mychka_components/ui-builder/default/pages/modal)

### DIFF
--- a/components/bl-modal-component/src/index.js
+++ b/components/bl-modal-component/src/index.js
@@ -31,19 +31,15 @@ export default function ModalComponent({ component, eventHandlers, pods, elRef }
     }
   };
 
-  if (!display) {
+  if (!display || !visibility) {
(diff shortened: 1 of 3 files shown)


components/bl-modal-component/src/index.js
```diff
@@ -31,17 +31,19 @@ export default function ModalComponent({ component, eventHandlers, pods, elRef }
     }
   };
 
-  if (!visibility) {
+  if (!display) {
```
v-excelsior, line 34:
```suggestion
  if (!display || !visibility) {
```

If I understand correct, in that case we could avoid using fragment

v-excelsior (changes requested):

PR #473, opened 2023-04-17
MARKET-1830 Tour Guide component

serhmel:
[UI Builder](https://develop.backendless.com/app/serhiy_melnychuk_components/ui-builder/TourGuide/pages/landing)
[Desktop Preview](https://genteelinstrument.backendless.app/api/files/ui-builder/containers/TourGuide/index.html?page=landing)

image
image

UI Builder Preview:
ui_builder_preview


components/bl-tour-guide-component/src/index.js
```diff
@@ -0,0 +1,111 @@
...
+  const [steps, setSteps] = useState(options);
+  const [helpers, setHelpers] = useState(null);
+  const [stepsKey, setStepsKey] = useState(0);
+
+  useEffect(() => {
+    setSteps(options);
+    setStepsKey(stepsKey + 1);
+  }, [options]);
+
+  const handleClickStart = event => {
```
v-excelsior, line 34: handleClickStart -> onStartClick

components/bl-tour-guide-component/src/index.js
```diff
@@ -0,0 +1,111 @@
...
+    event.preventDefault();
+
+    setRun(true);
+  };
+
+  const handleCallback = tourState => {
```
v-excelsior, line 40: onCallback

v-excelsior (changes requested):

components/bl-tour-guide-component/src/index.js
```diff
@@ -0,0 +1,111 @@
...
+    closeOnClickOverlay, scrolling, locale, scrollOffset, scrollDuration, scrollToFirstStep, spotlightClicks,
+  } = component;
+
+  const {
+    progressVisibility, skipButtonVisibility, overlayVisibility, backButtonVisibility, closeButtonVisibility,
+  } = component;
+
+  const {
+    spotlightPadding, beaconSize, arrowColor, backgroundColor, overlayColor, primaryColor, textColor, width,
+  } = component;
```
v-excelsior, line 20:
```suggestion
  const {
    classList, display, style, controlVisibility, controlLabel, options, continuous, autoStart, closeOnEscape,
    closeOnClickOverlay, scrolling, locale, scrollOffset, scrollDuration, scrollToFirstStep, spotlightClicks,
    progressVisibility, skipButtonVisibility, overlayVisibility, backButtonVisibility, closeButtonVisibility,
    spotlightPadding, beaconSize, arrowColor, backgroundColor, overlayColor, primaryColor, textColor, width,
  } = component;
```

Could we make use destruction once?

components/bl-tour-guide-component/src/index.js
```diff
@@ -0,0 +1,111 @@
...
+    if (status === STATUS.FINISHED || status === STATUS.SKIPPED || action === ACTIONS.RESET) {
+      setRun(false);
+    }
+
+    if (continuous && isChangeStepAction) {
+      setTimeout(() => {
+        setRun(false);
+        setRun(true);
+      });
+    }
```
v-excelsior, line 55: I'm not clearly understand why we use this approach, but didn't controll `stepIndex`

Especially looks strange 51-54 lines where we use a timeout

v-excelsior (changes requested):

components/bl-tour-guide-component/src/index.js
serhmel: This approach is not associated with stepIndex. Setting a stepIndex will turn Joyride into controlled mode, which will require additional logic that is currently being executed by the library.

According to the docs, need to set our running state to false, so we can restart if we click start again.:
```
    if (status === STATUS.FINISHED || status === STATUS.SKIPPED || action === ACTIONS.RESET) {
      setRun(false);
    }
```


As for this logic:
```
    const isNextStepAction = action === ACTIONS.NEXT && status !== STATUS.FINISHED;
    const isChangeStepAction = isNextStepAction || action === ACTIONS.PREV || action === ACTIONS.GO;

    if (continuous && isChangeStepAction) {
      setTimeout(() => {
        setRun(false);
        setRun(true);
      });
    }
```
The library has a bug with the tooltip appearance in some cases of using next, prev and goTo helpers. To fix this issue, we need to update the tour state.
We use timeout because we need to update the state after the current callback is completed.

v-excelsior: Is controlled mode requires a lot of logic to implement? If not - it will be good for the future.

If we leave this approach - please add comments on why we do it(also if the library has a known issue - add this link too)

That is good practice for all unobvious solutions.

PR #485, opened 2023-04-25
MARKET-1917 [Carousel] Strange component behaviour when there is only one image in JSON.

StasivPavlo:


components/bl-carousel-component/src/index.js
```diff
@@ -39,14 +39,14 @@ export default function Carousel({ component, eventHandlers }) {
   }, [imagesData]);
 
   useEffect(() => {
-    if (autoplayDelay && autoplay) {
+    if (autoplayDelay && autoplay && imagesList.length > 1) {
```
v-excelsior, line 42: `const isCarouserSwipeable = imagesList.length > 1 // isSwipeable // swipable // any other name that feat it`

and at 124-125 use this constant too

PR #501, opened 2023-05-04
MARKET-1682 [Transitions] Add the Guide for customers.

StasivPavlo:


components/bl-transitions-component/README.md
```diff
@@ -52,3 +52,15 @@ For example:
 <img alt="Close button logic" src="./example-images/close_button_logic.jpg" width="900" />
 
 In the preview, you now have the option to change the visibility of the component with animation by clicking on either the Close or Open button.
+
+### Adding component with dynamic content in transition
+It can come in handy when animating a block that requires some time to render completely. To good utilize the Collapse Top and Collapse Left effects, it's important to employ this guid.
```
v-excelsior, line 57: If you want to animate content that is rendered with a delay (for example, asynchronously loaded) follow the guide:

components/bl-transitions-component/README.md
```diff
@@ -52,3 +52,15 @@ For example:
...
+1. Firstly, you need activate it in the settings:
+
+> WARNING: After this action the component will not be displayed
```
v-excelsior, line 60: Where? Write about it.

```...will not be displayed in preview mode``` or something like that.

File: components/bl-transitions-component/README.md [hunk 3]
```diff
@@ -52,3 +52,15 @@ For example:
...
+
+<img alt="Close button logic" src="example-images/enable_dynamic_content.jpg" />
+
+2. Secondly, your block must have event that triggered when he full loaded. In him, we add component action "Set Content Loaded"
```
v-excelsior, line 64: Remove first and secondly words, you already have ordered list with numbers

v-excelsior: Also, what about rename actions `Set Content Loaded Transitions: <transition name>`  for `Set Content Loaded for: <transition name>`

Ask @snaqqqystas what he think about it, and if we decide to do it, please remake screenshots.

v-excelsior, line 64: How about:

> Run the action `Set Content Loaded` when your components are fully rendered. It could be done in `On ... Mount` handlers.

v-excelsior (changes requested):

components/bl-transitions-component/README.md
snaqqqystas: it's ok, but with it maybe change the effects raw and default effect to another one, - and if user will use the image without props, default effect normally work?

snaqqqystas: i've made another ticket for default effect changing.

components/bl-transitions-component/README.md
```diff
@@ -52,3 +53,16 @@ For example:
 <img alt="Close button logic" src="./example-images/close_button_logic.jpg" width="900" />
 
 In the preview, you now have the option to change the visibility of the component with animation by clicking on either the Close or Open button.
+
+### Adding component with dynamic content in transition
+If you want to animate content that is rendered with a delay (for example, asynchronously loaded) follow the guide:
+
+1. You need activate it in the settings:
```
v-excelsior, line 60:
```suggestion
1. You need to activate it in the settings:
```

components/bl-transitions-component/README.md
```diff
@@ -52,3 +53,16 @@ For example:
...
+### Adding component with dynamic content in transition
+If you want to animate content that is rendered with a delay (for example, asynchronously loaded) follow the guide:
+
+1. You need to activate it in the settings:
+
+> WARNING: After this action the component will not be displayed in preview mode
+
+<img alt="Close button logic" src="example-images/enable_dynamic_content.jpg" />
+
+2. Run the action `Set Content Loaded` when your components are fully rendered. It could be done in `On ... Mount handlers`.
```
v-excelsior, line 66:
```suggestion
2. Run the action `Set Content Loaded` when your components are fully rendered. It could be done in `On ... Mount` handlers.
```

components/bl-transitions-component/README.md
```diff
@@ -52,3 +53,16 @@ For example:
 <img alt="Close button logic" src="./example-images/close_button_logic.jpg" width="900" />
 
 In the preview, you now have the option to change the visibility of the component with animation by clicking on either the Close or Open button.
+
+### Adding component with dynamic content in transition
+If you want to animate content that is rendered with a delay (for example, asynchronously loaded) follow the guide:
+
+1. You need to activate it in the settings:
+
+> WARNING: After this action the component will not be displayed in preview mode
```
v-excelsior, line 62:
```suggestion
> 🔴 **IMPORTANT**: After this action the component will not be displayed in preview mode
```

v-excelsior (approved):

PR #502, opened 2023-05-04
MARKET-1700 [Transitions] The "Collapse top" effect works incorrectly.

StasivPavlo:


components/bl-transitions-component/src/collapse-left.js
```diff
@@ -18,17 +18,17 @@ export function CollapseLeft({ component, eventHandlers, transitionsContainerPod
...
   useEffect(() => {
     if (rootRef.current) {
-      hideElement(rootRef.current);
-
-      getWidthTimeout.current = setTimeout(() => {
-        rootRef.current.style.width = 'auto';
-        setWidth(rootRef.current.clientWidth);
-        rootRef.current.style.width = '0px';
-      }, 50);
+      if (isContentLoaded) {
```
v-excelsior, line 21: combine if-s

components/bl-transitions-component/src/collapse-left.js
```diff
@@ -18,17 +18,17 @@ export function CollapseLeft({ component, eventHandlers, transitionsContainerPod
 
   useEffect(() => {
```
v-excelsior, line 19:
```
  useEffect(() => {
    let getWidthTimeout
    
    if (rootRef.current && isContentLoaded) {
        getWidthTimeout = setTimeout(() => {
          rootRef.current.style.width = 'auto';
          setWidth(rootRef.current.clientWidth);
          rootRef.current.style.width = '0px';
        }, 50);
    }

    return () => getWidthTimeout && clearTimeout(getWidthTimeout);
  }, [rootRef, isContentLoaded]);
  ```

File: components/bl-transitions-component/src/collapse-left.js [hunk 3]
```diff
@@ -41,7 +41,7 @@ export function CollapseLeft({ component, eventHandlers, transitionsContainerPod
       <div
         ref={ rootRef }
         className={ cn('transition', variants, { [variants + '--active']: isTransition }) }
-        style={{ ...style, transitionDuration: duration + 'ms' }}>
+        style={{ ...style, transitionDuration: duration + 'ms', position: 'absolute', zIndex: -1, opacity: 0 }}>
```
v-excelsior, line 44: Do we need it in JS styles?

```
position: 'absolute', zIndex: -1, opacity: 0
```

Could it come from the styles file?

components/bl-transitions-component/src/collapse-top.js
```diff
@@ -41,7 +39,7 @@ export function CollapseTop({ component, eventHandlers, transitionsContainerPod,
       <div
         ref={ rootRef }
         className={ cn('transition', variants, { [variants + '--active']: isTransition }) }
-        style={{ ...style, transitionDuration: duration + 'ms' }}>
+        style={{ ...style, transitionDuration: duration + 'ms', position: 'absolute', zIndex: -1, opacity: 0 }}>
```
v-excelsior, line 42: Same here, could it come from styles?

If yes - don't forget about variables.

v-excelsior: If no - create constant showElementStyles / hideElementStyles(or something like that) and use destruction here.

v-excelsior (changes requested):

components/bl-transitions-component/src/collapse-top.js
StasivPavlo: > If yes - don't forget about variables.

I think that there is no need to add variables in the styles, because if you change these properties, the animation may not work correctly

PR #516, opened 2023-05-16
MARKET-1964 [Barcode] "Mouse over/out" events works wrong

serhmel:


components/bl-barcode-component/styles/index.less
```diff
@@ -5,7 +5,11 @@
   display: flex;
   width: @bl-customComponent-barcode-width;
 
-  img {
+  svg {
     width: @bl-customComponent-barcode-image-width;
+
```
v-excelsior, line 10: I assume we should move mouse event to the div that contain label and svg

v-excelsior (changes requested):

components/bl-barcode-component/styles/index.less
serhmel: I have nothing against it, I just want to clarify that the label is placed inside the svg. The "div" element is just a wrapper.
Do you mean to do like this?

```
    <div
      ref={ elRef }
      className={ cn('bl-customComponent-barcode', classList) }
      style={ style }
      onClick={ event => onClick({ event }) }
      onMouseOver={ event => onMouseOver({ event }) }
      onMouseOut={ event => onMouseOut({ event }) }>
      <svg ref={ barcodeRef }/>
    </div>
```

PR #518, opened 2023-05-17
MARKET-1951 [Transitions] Component can't get the % dimension of the element inside.

StasivPavlo:


components/bl-transitions-component/src/collapse-left.js
```diff
@@ -10,26 +10,34 @@ export function CollapseLeft(props) {
   const { onMounted, onUnmounted, onEndAnimation, onStartAnimation } = eventHandlers;
 
   const [width, setWidth] = useState(0);
+  const [initWidth, setInitWidth] = useState('');
 
   const rootRef = useRef();
+  const [element, setElement] = useState({ current: null });
```
v-excelsior, line 16: Why we need this `{ current: null }`?

components/bl-transitions-component/src/collapse-left.js
```diff
@@ -10,26 +10,34 @@ export function CollapseLeft(props) {
...
-    if (rootRef.current && isContentLoaded && isImagesLoaded) {
-      getWidthTimeout = setTimeout(() => {
-        rootRef.current.style.width = 'auto';
-        setWidth(rootRef.current.clientWidth);
-        rootRef.current.style.width = '0px';
-      }, 50);
+    if (element.current && isContentLoaded && isImagesLoaded) {
+      const size = element.current.clientWidth;
+
+      getWidthTimeout = setTimeout(() => setWidth(size), 50);
```
v-excelsior, line 34: Why timeout is `50`?

components/bl-transitions-component/src/others.js
```diff
@@ -45,17 +52,25 @@ export const Others = props => {
     }
   }, [isTransition]);
 
+  useEffect(() => {
+    if (element.current) {
+      element.current.classList.add('transition', variants);
+
+      if (display && isTransition) {
+        element.current.classList.add(variants + '--active');
+        element.current.style.transitionDuration = duration + 'ms';
```
v-excelsior, line 61: function onOpen/onHide or something like that

components/bl-transitions-component/src/collapse-left.js
```diff
@@ -43,14 +51,24 @@ export function CollapseLeft(props) {
     }
   }, [isTransition]);
 
+  useEffect(() => {
+    if (element.current) {
+      element.current.classList.add('transition', variants);
```
v-excelsior, line 56: variants is plural?

components/bl-transitions-component/src/collapse-left.js
```diff
@@ -43,14 +51,24 @@ export function CollapseLeft(props) {
     }
   }, [isTransition]);
 
+  useEffect(() => {
```
v-excelsior, line 54: Could you add some levels of abstraction where it possible?

components/bl-transitions-component/src/collapse-top.js
```diff
@@ -10,26 +10,34 @@ export function CollapseTop(props) {
...
-  const isTransition = useTransition(rootRef, display, duration, height, 'height', setIsAuto, onEndAnimation);
-  const isImagesLoaded = useImageLoad(rootRef, dynamicContent);
+  const isImagesLoaded = useImageLoad(element, dynamicContent);
+  const isTransition = useTransition(element, display, duration, initHeight, height, 'height', onEndAnimation);
+
+  useEffect(() => {
+    if (rootRef.current && !element.current) {
+      setElement({ current: rootRef.current.firstElementChild });
+      setInitHeight(rootRef.current.firstElementChild.style.height);
+    }
```
v-excelsior, line 25: For example

```suggestion
    const readyToInitialTransition = rootRef.current && !element.current // or what we do here, i don't clearly undestand

    if (readyToInitialTransition) {
      setElement({ current: rootRef.current.firstElementChild });
      setInitHeight(rootRef.current.firstElementChild.style.height);
    }
```

This little things makes code more readable

v-excelsior (changes requested):

components/bl-transitions-component/src/collapse-left.js
```diff
@@ -6,29 +6,32 @@ const { cn } = BackendlessUI.CSSUtils;
...
 
   const isImagesLoaded = useImageLoad(element, dynamicContent);
   const isTransition = useTransition(element, display, duration, initWidth, width, 'width', onEndAnimation);
 
   useEffect(() => {
-    if (rootRef.current && !element.current) {
+    const readyToInitialTransition = rootRef.current && !element.current;
+
+    if (readyToInitialTransition) {
       setElement({ current: rootRef.current.firstElementChild });
```
v-excelsior, line 25: Same question as here https://github.com/Backendless/ui-builder-library/pull/518#discussion_r1196790726

Why we need property current?

components/bl-transitions-component/src/collapse-left.js
v-excelsior: Still unanswered

components/bl-transitions-component/src/collapse-left.js
```diff
@@ -53,7 +56,7 @@ export function CollapseLeft(props) {
 
   useEffect(() => {
     if (element.current) {
-      element.current.classList.add('transition', variants);
+      element.current.classList.add('transition', variant);
```
v-excelsior, line 59: What does `element`? 'element` is pretty much the same as `data`, it could be completely everything and doesn't describe anything at the same time. Please, choose a meaningful name or explain why we choose `element` name.

components/bl-transitions-component/src/index.js
```diff
@@ -14,13 +14,13 @@ const transitionsViews = {
...
 
 export default function Transitions({ component, eventHandlers, pods }) {
-  const { variants, display, duration, dynamicContent } = component;
+  const { variant, display, duration, dynamicContent } = component;
   const { onEndAnimation } = eventHandlers;
 
   const [isContentLoaded, setIsContentLoaded] = useState(!dynamicContent);
 
   const transitionsContainerPod = pods['transitionsContainer'];
-  const Transitions = transitionsViews[variants];
```
v-excelsior, line 23: transitionsViews - TransitionView or TransitionViewMap
also, I thing correct spelling is `transition views`(without s after transition)

Transitions -> Transition or TransitionComponent

components/bl-transitions-component/src/others.js
```diff
@@ -54,23 +56,30 @@ export const Others = props => {
...
 
   return (
-    <div ref={ rootRef } className={ cn('bl-customComponent-transitions', variants, classList) } style={ style }>
+    <div ref={ rootRef } className={ cn('bl-customComponent-transitions', variant, classList) } style={ style }>
       { transitionsContainerPod.render() }
     </div>
   );
 };
+
+const onOpen = (element, variant, duration) => {
```
v-excelsior, line 77: Here shouldn't be current proiperty, function should work with clear html element.

v-excelsior (changes requested):

components/bl-transitions-component/src/collapse-left.js
StasivPavlo: there is a possibility that the component appears already with the final height and the transition is not visible. When I add the timeout to set the final height, I don't see this possibility. It doesn't have to be 50, I put 50 with a margin because for example if you set no delay or less than 5ms you might have this possibility

StasivPavlo: Also I detected this possibility only when component rendered at on time with page

v-excelsior: What is the reason for this "possibility"?

Each time when you use the timeout with some not reasonable value(50 in your example) it says that we trying to make a workaround in a very imperative way. Why it's bad? For example, I opened the page on a slow device, does 50 still works? Or if render of children's content takes more than 50ms?

Sometimes it makes sense, but at least I want to know we have an understanding of the nature of this behavior.

components/bl-transitions-component/src/index.js
v-excelsior: You miss that: 

transitionsViews -> TransitionView or TransitionViewMap

v-excelsior (changes requested):

PR #531, opened 2023-05-23
MARKET-1715 [Dashlet] Width Logic handler and Height Logic handler do not work

StasivPavlo:


components/bl-dashlet-component/src/dashlet.js
```diff
@@ -29,6 +30,14 @@ export function Dashlet(props) {
     return {};
   }, [height, width, isOpen]);
 
+  useEffect(() => {
+    setResizeMaxHeight(maxHeight);
+  }, [maxHeight]);
+
+  useEffect(() => {
+    setResizeMaxWidth(maxWidth);
+  }, [maxWidth]);
```
v-excelsior, line 39: both could be one lined

components/bl-dashlet-component/src/dashlet.js
```diff
@@ -51,14 +60,14 @@ export function Dashlet(props) {
     onDrag         : handleDrag,
     rootRef        : rootRef,
     initialPosition: position,
-    dragging
+    dragging,
   });
 
   const onResizeStop = useCallback((e, data) => {
```
v-excelsior, line 66:
```suggestion
  const onResizeStop = useCallback((e, {size: {height, width}}) => {
```

v-excelsior (changes requested):

components/bl-dashlet-component/src/dashlet.js
```diff
@@ -29,6 +30,10 @@ export function Dashlet(props) {
     return {};
   }, [height, width, isOpen]);
 
+  useEffect(() => setResizeMaxHeight(maxHeight), [maxHeight]);
+
```
v-excelsior, line 34:
```suggestion
```

PR #534, opened 2023-05-25
BKNDLSS-32562 [Mapbox Component] Add polygon outline.

addKiwi:
image


components/bl-mapbox-component/src/helpers.js
```diff
@@ -217,17 +220,24 @@ const createPopup = (polygon, mapRef, onPolygonClick, map) => {
...
 
-const addPolygons = (mapRef, polygonsArray, onPolygonClick, map) => {
+const addPolygons = (mapRef, polygonsArray, onPolygonClick, map, polygonOutline, polygonOutlineColor,
+                     polygonOutlineWidth) => {
   polygonsArray.forEach(polygon => {
     const { id, coordinates, color, opacity } = polygon;
+    const verifiedCoordinates = [...coordinates];
+
+    if (coordinates[0][0] !== coordinates[coordinates.length - 1][0]
+      && coordinates[0][1] !== coordinates[coordinates.length - 1][1]) {
```
v-excelsior, line 230: create a function for that condition to understand what going on here

components/bl-mapbox-component/src/helpers.js
```diff
@@ -243,21 +253,35 @@ const addPolygons = (mapRef, polygonsArray, onPolygonClick, map) => {
...
+      map.addLayer({
+        id    : `${ id }-outline`,
+        type  : 'line',
+        source: id,
+        layout: {},
+        paint : {
+          'line-color': polygonOutlineColor,
+          'line-width': polygonOutlineWidth
+        }
+      });
```
v-excelsior, line 266: MapController class could have a method that hide this logic from the component.
Here it could looks like

`map.addPolygonOutline(id, polygonOutlineColor,polygonOutlineWidth)`

components/bl-mapbox-component/README.md
```diff
@@ -10,24 +10,27 @@ The component covers basic features of the library and allows you in a couple cl
...
+| Polygon Outline Color | *Color*                                                                                                                         | '#000000'                            |                | NO           | YES        | Sets the polygon outline color.                                                                                                                                                                         |
+| Polygon Outline Width | *Number*                                                                                                                        | 2                                    |                | NO           | YES        | Setts the polygon outline width.                                                                                                                                                                        |
```
v-excelsior, line 19: Could it be added to polygon signature? In that case we have an ability to create different polygons.

v-excelsior (changes requested):

PR #559, opened 2023-06-05
MARKET-2005 Parallax isn't scrolled

StasivPavlo:


components/bl-parallax-component/src/index.js
```diff
@@ -47,18 +47,40 @@ const useAnimation = (backdropRef, containerRef, strength, display) => {
...
+
+      scrollElement.addEventListener('scroll', animate);
       window.addEventListener('resize', animate, false);
     }
 
     return () => {
-      document.removeEventListener('scroll', animate);
+      scrollElement.removeEventListener('scroll', animate);
       window.removeEventListener('resize', animate, false);
     };
```
v-excelsior, line 66: Return should be moved info `if` condition, isn't it?

PR #582, opened 2023-06-13
MARKET-2031 [Totop] The scroll up works from 2nd click right after full scroll down.

serhmel:


components/bl-totop-component/src/index.js
```diff
@@ -37,7 +39,7 @@ export default function TotopComponent({ component, elRef }) {
   const handleScroll = useCallback(() => {
     const offsetTop = element?.el?.offsetTop || 0;
 
-    window.scrollTo({ top: offsetTop + offset, behavior: 'smooth' });
```
v-excelsior, line 40: SMOOTH_SCROLL_BEHAVIOR -> redundant constant as it browser api and used only once. Also 300ms delay for  this action is looking not as good solution. 

What the root problem of the issue? Element is not ready for the moment when user click?

PR #585, opened 2023-06-13
MARKET-2027 [Transitions] Add the "Is open" Checkbox to settings.

StasivPavlo:


components/bl-transitions-component/component.json
```diff
@@ -8,6 +8,18 @@
...
+      "type": "checkbox",
+      "name": "isOpen",
+      "label": "Is Open",
+      "showInSettings": true,
+      "defaultValue": true,
+      "hasLogicHandler": true,
+      "handlerId": "isOpenLogic",
+      "handlerLabel": "Is Open Logic",
+      "dataBinding": true,
+      "handlerDescription": "This handler allows for determining if the component is open or closed."
```
v-excelsior, line 21: Tooltips?

components/bl-transitions-component/src/helpers.js
```diff
@@ -25,16 +25,12 @@ export const useTransition = (
       if(readyToStartTransition) {
         setPodElementDimension(podElement['client' + dimensionName]);
         setIsTakenMeasurements(true);
+
+        transitionRef.current.style[dimensionName.toLowerCase()] = '0px';
```
v-excelsior, line 29: `dimensionName.toLowerCase()` looks dirty. Also as `'client' + dimensionName`

Could we get rid of it?

v-excelsior (changes requested):

PR #588, opened 2023-06-15
MARKET-2046 [Marquee] Add new helpful events.

serhmel:


components/bl-marquee-component/src/index.js
```diff
@@ -70,7 +70,13 @@ export default function MarqueeComponent({ component, elRef, eventHandlers, pods
   }
 
   return (
-    <div ref={ elRef } className={ cn('bl-customComponent-marquee', ...classList) } style={ styles }>
+    <div
+      ref={ elRef }
+      className={ cn('bl-customComponent-marquee', ...classList) }
```
v-excelsior, line 75: No need to use `...` with cn

v-excelsior, line 75:
```suggestion
      className={ cn('bl-customComponent-marquee', classList) }
```

PR #600, opened 2023-06-19
MARKET-2059 [Dashlet] Error "Cannot read properties of undefined (reading 'length')" appeared after running preview with handler dragging logic = false

StasivPavlo:


components/bl-dashlet-component/src/helpers.js
```diff
@@ -11,75 +11,79 @@ export const ContextBlockItemTypes = {
...
+        handleMouseMove.cancel();
 
-      document.removeEventListener('mousemove', handleMouseMove);
-      document.removeEventListener('mouseup', handleMouseUp);
-    };
+        document.removeEventListener('mousemove', handleMouseMove);
+        document.removeEventListener('mouseup', handleMouseUp);
+      };
+    }
   }, [pressed, onDrag]);
```
v-excelsior, line 81: `dragging` in deps is not needed?

PR #612, opened 2023-06-21
MARKET-2038 [Audio Recorder] Downloaded files doesn't work.

addKiwi:


components/bl-audio-recorder/src/helpers.js
```diff
@@ -5,7 +5,7 @@ export const download = (blob, fileName, fileType) => {
 
   link.style.display = 'none';
   link.href = URL.createObjectURL(blob);
-  link.download = `${ fileName }.${ fileType }`;
+  link.download = `${ fileName }.${ fileType.toLowerCase() }`;
```
v-excelsior, line 8: Why do we need it?

addKiwi: To avoid extensions in upper case. I'm not sure but it could be a problem for some audio players.

PR #613, opened 2023-06-21
MARKET-1973 Data Grid component fails

Saharij:


components/bl-data-grid/src/utils.js
```diff
@@ -0,0 +1,33 @@
...
+
+    return parentObj;
+  });
+}
+
+function flattenObject(parent, nestedObj) {
+  const parentObj = parent;
+
+  for (const key in nestedObj) {
+    if (typeof nestedObj[key] === 'object' && nestedObj[key] !== null && !Array.isArray(nestedObj[key])) {
```
v-excelsior, line 25: utility function for that 'isObject' or something like it

components/bl-data-grid/src/utils.js
```diff
@@ -0,0 +1,33 @@
+export function validateRows(data) {
+  if (!data) {
+    return [];
+  }
+
+  return data.map(item => {
+    const parentObj = {};
+
+    for (let key in item) {
+      if (typeof item[key] !== 'object' || item[key] === null || Array.isArray(item[key])) {
```
v-excelsior, line 10: I'm not sure, but it could be `!isObject(v)` 
see comment below

PR #638, opened 2023-07-04
MARKET-2070 [Dashlet] Dashlet component overlaps another component 

StasivPavlo:


components/bl-dashlet/styles/index.less
```diff
@@ -41,8 +41,10 @@
     cursor: pointer;
   }
 
-  z-index: 1;
-  position: absolute;
+  &.dragging {
+    z-index: 1;
```
v-excelsior, line 45: I assume there should be variable

components/bl-dashlet/src/index.js
```diff
@@ -46,7 +46,7 @@ export default function DashletComponent({ component, eventHandlers, pods, insta
   return (
     <div
       ref={ rootRef }
-      className={ cn('bl-customComponent-dashlet', classList) }
+      className={ cn('bl-customComponent-dashlet', classList, { dragging }) }
```
v-excelsior, line 49: Are use sure that is `dragging`? Dragging - is a process, if we want to say: dashlet could be dragged - there should be `draggable`

PR #643, opened 2023-07-05
MARKET-2087 Upgrade lib Dropzone to Files UI in the Dropzone Custom UI Component

serhmel:


components/bl-dropzone/src/index.js
```diff
@@ -38,50 +36,44 @@ export default function DropzoneComponent({ component, eventHandlers }) {
...
-        onClose={ () => handleSee(undefined) }
-      />
-      <VideoPreview
-        videoSrc={ videoSrc }
-        openVideo={ videoSrc }
-        onClose={ () => handleWatch(undefined) }
-        controls
-        autoplay
-      />
+      <FullScreen open={ imageSrc !== undefined } onClose={ handleSee }>
```
v-excelsior, line 71: Wondering about why `imageSrc: false` is positive variant here.

v-excelsior (changes requested):

components/bl-dropzone/src/index.js
serhmel: The `imageSrc` value cannot be `false`.
By default, it is `undefined`, when clicking the preview button, we set it to `imageSource` or `videoSource`, when clicking to close the preview, we set `imageSrc` to `undefined`.

v-excelsior: If it can't be any kind of falsy values except `undefined` why we couldn't use
`open={imageSrc}` or `open={!!imageSrc}`

PR #657, opened 2023-07-13
BKNDLSS-33088 *Support*[Pagination] Add the useful Usage block to the Readme file.

StasivPavlo:


components/bl-pagination/README.md
```diff
@@ -10,16 +10,16 @@ The component based on external [pagination](https://mui.com/material-ui/react-p
...
+| Variant: `variant`                                       | Select [Normal: `normal`, Outline: `outline`, Rounded: `rounded`, Rounded-Outline: `rounded-outline`] | Normal: `normal` |                     | NO           | YES        | Allows to determine the variant of pagination style                        |
+| Size: `size`                                             | Select [Small: `small`, Normal: `normal`, Large: `large`]                                             | Normal: `normal` |                     | NO           | YES        | Allows to determine the size for pagination                                |
+| Is Next Button Visible: `isNextButtonVisible`            | Checkbox                                                                                              | `true`           |                     | NO           | YES        | Allows to determine the visible or not visible "Next" button               |
```
v-excelsior, line 19: That should be names as setting `Is Next Button Visible` -  `Next Button`

Same for other

components/bl-pagination/README.md
```diff
@@ -60,3 +60,58 @@ The component based on external [pagination](https://mui.com/material-ui/react-p
...
+
+### After we make Row Data Logic for Data Grid
+
+Make variable “pageSize” where we will save how many rows must be in one page
+
+![Make variable](./example-images/make-variable-page-size.png)
+
+Now we need to calculate how many pages we have. For that, we use the block for getting the count of rows in our table and then divide this count by the count of rows on one page(pageSize). This result we round up for integer number:
+
+![Calculate pages](./example-images/calculate-page-count.png)
```
v-excelsior, line 89: Check in devtools how many times this request is run. There should be a single request, at first render.

components/bl-pagination/README.md
```diff
@@ -60,3 +60,58 @@ The component based on external [pagination](https://mui.com/material-ui/react-p
...
+
+After we set this result to “Page Data”:
+
+![Set result](./example-images/set-page-count-to-page-data.png)
+
+Now we load the table's data in “return” and calculate page offset, where “currentPage” we will get from the Pagination event “On Page Change”:
+
```
v-excelsior, line 96: Same here.

Forget for requests in `... Logic` handlers.

You should load pages at `On Page Change request` and store them into some variable in Page Data(for example) that will be used by Data Grid

components/bl-pagination/src/index.js
```diff
@@ -54,7 +55,7 @@ export default function Pagination({ component, eventHandlers }) {
 
         { isFirstPageButtonVisible && (
           <FirstPageButton
-            onGoFirst={ onGoFirst }
+            onGoFirst={ onGoFirst.hasLogic ? onGoFirst : component.goFirstPage }
```
v-excelsior, line 58: We skip current page handling if there is user handler? Why?

v-excelsior (changes requested):

components/bl-pagination/src/index.js
StasivPavlo: If the user will have some logic in for example On Go First he needs to use the action Go to First Page. But when his logic is empty we use this action by "default"

components/bl-pagination/src/index.js
```diff
@@ -40,6 +41,14 @@ export default function Pagination({ component, eventHandlers }) {
     setCurrentPage(pageCount);
   };
 
+  useEffect(() => {
+    onMounted();
+
+    return () => {
+      onUnmount();
+    };
```
v-excelsior, line 49:
```
 return () => onUnmount();
 ```

File: components/bl-pagination/README.md [hunk 6]
```diff
@@ -10,26 +10,28 @@ The component based on external [pagination](https://mui.com/material-ui/react-p
...
+| On Go Next        | when the user clicks on next button                                                                                                                                                                             |                       |
+| On Go Back        | when the user clicks on back button                                                                                                                                                                             |                       |
+| On Go First       | when the user clicks on first-page button                                                                                                                                                                       |                       |
+| On Go Last        | when the user clicks on last-page button                                                                                                                                                                        |                       |
```
v-excelsior, line 34: On Go Next -> On Next Button Click
same for others

components/bl-pagination/README.md
```diff
@@ -60,3 +62,48 @@ The component based on external [pagination](https://mui.com/material-ui/react-p
...
+
+![Calculate pages](./example-images/calculate-page-count.png)
+
+Add “pageCount” from “Page Data” to data binding for Page Count Logic:
+
+![Add pageCount](./example-images/add-page-count-to-data-binding.png)
+
+Now we load the table's data in Pagination On Page Change event and calculate page offset and set it to “Page Data”:
+
+![Load data](./example-images/get-data-from-table.jpg)
```
v-excelsior, line 95: Make sense to group images. We could do it in two ways.
First - example prefix.
`usage-<what happens on pickture>.jpg`

Or we could create a separate directory for every example(i'f every example contains a lot of images).

It helps us to not mix up different images and makes pictures like `result` more obvious.

v-excelsior (changes requested):

PR #663, opened 2023-07-17
BKNDLSS-33116 [Simple Modal] Add possibility to use HTML at content property

StasivPavlo:


components/bl-simple-modal/src/subcomponents.js
```diff
@@ -36,9 +36,7 @@ export function Container(props) {
   return (
     <div className="simple-modal__container">
       { content && (
-        <p className="simple-modal__text">
-          { content }
-        </p>
+        <p className="simple-modal__text" dangerouslySetInnerHTML={{ __html: content }}></p>
```
v-excelsior, line 39: I'm not sure if it complete solution as far as it is open for XSS.

There are three variants of what we can do:

1) Add a checkbox that will clean content from unexpected JS code.
2) Notify the user about a possible XSS attack and do nothing.
3) Forcibly clean HTML and add exclamation in the docs.

I prefer the third variant but want to ask you to think about the pros and cons. 

Code that could clean HTML - already written, so you could just ask me when it will be needed.

PR #672, opened 2023-07-20
MARKET-1749 [Social Login Buttons] Rename the "Googleplus" to the "Google".

Saharij:


components/bl-social-login-buttons/src/index.js
```diff
@@ -2,6 +2,7 @@ import { Button } from './button';
 
 const { cn } = BackendlessUI.CSSUtils
 
+const GOOGLEPLUS = 'googleplus';
 const providersList = ['googleplus', 'facebook', 'twitter', 'linkedin', 'github'];
```
v-excelsior, line 6:
```
const ProvidersMap = {PROVIDER_NAME: 'provider-value'}
const ProvidersList = Object.values(ProvidersMap)

const ProviderLabels = {
  [ProvidersMap.PROVIDER_NAME]: 'provider label'
}

...

buttonLabel={ `Connect with ${ ProviderLabels[providerCode]` }

```

v-excelsior (changes requested):

PR #682, opened 2023-07-25
BKNDLSS-33207 [Mapbox component] Missing context block for On Marker Click Event

addKiwi:


components/bl-mapbox/README.md
```diff
@@ -34,7 +34,7 @@ The component covers basic features of the library and allows you in a couple cl
 | Name                       | Triggers                               | Context Blocks                                                                                                                     |
 |----------------------------|----------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|
 | On Click                   | when a user click on map               | Coordinates: `{lat: Number, lng: Number}`                                                                                          |
-| On Marker Click            | when a user click on a marker          | Coordinates: `{lat: Number, lng: Number}` <br/> Description: `String`                                                              |
+| On Marker Click            | when a user click on a marker          | Coordinates: `{lat: Number, lng: Number}` <br/> Description: `String` <br/> Marker: `String or Number or Boolean or Object`        |
```
v-excelsior, line 37: or -> `|`

components/bl-mapbox/src/helpers.js
```diff
@@ -199,7 +199,7 @@ export const useMarkers = (markers, mapRef, onMarkerClick) => {
         popup.on('open', () => {
           const coordinates = { lat: markerItem.coordinates.lat, lng: markerItem.coordinates.lng };
 
-          onMarkerClick({ coordinates, description: description || '' });
+          onMarkerClick({ coordinates, description: description || '', markerData });
```
v-excelsior, line 202: Why do we need to do this with `coordinates`?

Can we get them from markerItem and pass here?

components/bl-mapbox/src/helpers.js
```diff
@@ -188,7 +188,7 @@ export const useMarkers = (markers, mapRef, onMarkerClick) => {
       setMarkersArray([]);
```
v-excelsior, line 188: Change useEffect + useState for useMemo

Ask Pavlo for isDataChanged function or write me, and recreate only new markers. As I see marker has `update` method.

components/bl-mapbox/component.json
```diff
@@ -204,6 +204,10 @@
         {
           "id": "description",
           "label": "Description"
+        },
+        {
+          "id": "markerData",
```
v-excelsior, line 209: Seems strange when Marker entity conteint `markerData` property. That should be `data` or something that means additional information inside market

components/bl-mapbox/src/helpers.js
```diff
@@ -199,7 +199,7 @@ export const useMarkers = (markers, mapRef, onMarkerClick) => {
         popup.on('open', () => {
           const coordinates = { lat: markerItem.coordinates.lat, lng: markerItem.coordinates.lng };
 
-          onMarkerClick({ coordinates, description: description || '' });
+          onMarkerClick({ coordinates, description: description || '', markerData });
         });
 
         if (description) {
```
v-excelsior, line 205: logic below

```
    if (polygon.outlineWidth > 0) {
        map.removeLayer(`${ polygon.id }-outline`);
      }
      map.removeLayer(`${ polygon.id }-layer`);
      map.removeSource(polygon.id);
      ```
      
Should be done inside one Map class method that accept `polygon`

Functions that already created is useless(removeLayer, removeSource)

File: components/bl-mapbox/src/helpers.js (hunk 3 above)
Comment (v-excelsior) [reply components/bl-mapbox/src/helpers.js:188]: Also take a loot at every `new` and ask yourself "is it correct to create new element in this situation or we could update existed?"

**Review by v-excelsior — CHANGES_REQUESTED** [review]

File: components/bl-mapbox/src/helpers.js [hunk 6]
```diff
@@ -177,51 +177,61 @@ export const initMapboxLibrary = (mapRef, mapContainerRef, component, eventHandl
...
+const removeMarkers = markers => {
+  if (markers?.length) {
+    markers.forEach(marker => {
+      marker.remove();
+    });
+  }
+};
+
+const changeMarkerColor = marker => {
+  marker.getElement().childNodes[0].classList.toggle('mapbox-marker-active');
```
v-excelsior, line 228: Why do we need `.childNodes[0]`. We can't set class on root marker element?

components/bl-mapbox/styles/index.less
```diff
@@ -4,6 +4,7 @@
 
 @bl-customComponent-mapbox-height: 400px;
 @bl-customComponent-mapbox-width: 600px;
+@bl-customComponent-mapbox-marker-click-color: red;
```
v-excelsior, line 7: click-color -> active-color

And by default it should be the same as not active marker or 'unset'

components/bl-mapbox/src/helpers.js
v-excelsior, line 228: Not it looks in style like

`bl-customComponent-mapbox mapbox-marker-active`

We don't need to repeat 'mapbox'.

Also if it is possible to remove childNodes from code higher, it could be `marker-root`

Last but not least - active is modification, so better to use another separation(double dash) -> `marker-root--active`

v-excelsior (changes requested):

PR #689, opened 2023-07-27
MARKET-2130 [PDF Viewer] Add a guide for using external links to PDF files to bypass CORS

StasivPavlo:


components/bl-pdf-viewer/README.md
```diff
@@ -139,3 +139,30 @@ The component allows you to add a pdf viewer to your UI-Builder application
 @bl-customComponent-pdfViewer-no-data-border-width: 100%;
 @bl-customComponent-pdfViewer-no-data-icon-fill: #fff;
 ```
+
+## <a id="external-link-guide"></a> Usage external link to PDF file
+Unfortunately, we are unable to resolve the issue with external links to PDF files due to the [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)
+policy on the server where the PDF file is stored. But we can suggest trying to bypass CORS. The idea is to send a request
+to a PDF file through a Cloud Code >> Services and use the File API to add the file to our file system, then transfer
```
v-excelsior, line 146: > Unfortunately, we are unable to resolve

Sound like it's our mistake or disadvantage - but it's not on our side. So it will be more correct to change the apologetic tone to advicing/explanation.

For example:

> If you facing ... blah-blan ... you could do next things/steps:

PR #696, opened 2023-08-02
MARKET-2136 [Popover] Add tooltips to the component settings.

StasivPavlo:


components/bl-popover/component.json
```diff
@@ -34,23 +34,26 @@
...
     },
     {
       "type": "number",
       "name": "delayMouseOut",
       "label": "Delay Mouse Out",
       "showInSettings": true,
       "defaultValue": "0",
-      "hasLogicHandler": false
+      "hasLogicHandler": false,
+      "settingTooltip": "Allows to determine the delay for On Mouse Out event"
```
v-excelsior, line 56: We could simplify it using expressions like "Enter Delay" and "Exit Delay"

v-excelsior: I'm not only about tooltips, but about labels and property name too

Also we should rename `Tooltip.js` file -> `tooltip.js`

And update example screenshots, that haven't this two handlers.

PR #704, opened 2023-08-09
BKNDLSS-33378 [Leaflet map] App crash after adding the JSON marker in settings.

addKiwi:


components/bl-leaflet-map/src/helpers/map.js
```diff
@@ -111,7 +111,7 @@ export function useMarkers(markers, icon, map, eventHandlers) {
   useEffect(() => {
     clearOldMarkers(markersArray);
 
-    if (markers) {
+    if (markers && map) {
```
v-excelsior, line 114:
```suggestion
    if (map && markers) {
```

components/bl-leaflet-map/src/helpers/map.js
```diff
@@ -121,7 +121,7 @@ export function useMarkers(markers, icon, map, eventHandlers) {
           .bindPopup(description);
       });
     }
-  }, [markers]);
+  }, [markers, map]);
```
v-excelsior, line 124:
```suggestion
  }, [map, markers]);
```

PR #706, opened 2023-08-15
MARKET-2139 [Avatar] Pointer and active UI reaction on the "Read only" mode empty image.

serhmel:


components/bl-avatar/src/index.js
```diff
@@ -1,5 +1,6 @@
 import { useEffect, useRef, useState } from 'react';
 
+import fallbackImage from './assets/fallback-image.jpg';
```
v-excelsior, line 3: Looks like constant - FALLBACK_IMAGE

PR #730, opened 2023-08-23
MARKET-2095 [Data Grid] Add useful sorting actions for columns.

Saharij:


components/bl-data-grid/component.json
```diff
@@ -271,6 +271,30 @@
...
+      ]
+    },
+    {
+      "id": "sortByColIdDesc",
+      "label": "Sort By Column Id Desc in",
+      "inputs": [
+        {
+          "id": "columnId",
+          "label": "Column Id"
+        }
```
v-excelsior, line 292: should be `sortBy({columnId, direction})`

we doesn't need both actions

Also, could we add starting sorting? Or we already have it?

Saharij: Regarding default column sorting, I found a simpler way to do it

Screenshot 2023-08-25 at 00 40 37

In this way, the customer can choose the desired column, as well as the sorting direction.

I propose this variant because I could not implement as we discussed at the call. There were problems with the ref.

So, if we choose the variant I'm suggesting, the customers need to know somehow that there is an option to specify a default column sorting.

PR #733, opened 2023-08-23
MARKET-2163 [Drawer] Default Open Logic isn't work

serhmel:


components/bl-drawer/src/helpers.js
```diff
@@ -51,10 +53,8 @@ export function useDrawerActions(component, eventHandlers, drawerContentRef, dra
   }, [animationDuration, drawerContainerRef, drawerContentRef, onClose, placement, visibility]);
 
   useEffect(() => {
-    if (defaultOpen) {
-      setVisibility(true);
-    }
-  }, [])
+    setVisibility(defaultOpen);
+  }, [defaultOpen]);
```
v-excelsior, line 57: For now it looks not like `defaultOpen` but like one more variation of visibility. What if defaultOpen will be changed dynamically from handler?

PR #750, opened 2023-08-31
MARKET-2177 [Dashlet] Component position reset after clicking on the Menu items button

StasivPavlo:


components/bl-dashlet/src/helpers.js
```diff
@@ -10,10 +10,10 @@ export const ContextMenuItemTypes = {
   ACTION: 'action',
 };
 
-export const useDraggable = ({ onDrag, rootRef, initialPosition, draggable }) => {
+export const useDraggable = ({ onDrag, rootRef, initialPosition, draggable, setPosition }) => {
   const [pressed, setPressed] = useState(false);
 
-  const position = useRef({ x: initialPosition.x, y: initialPosition.y });
+  const position = useMemo(() => ({ current: { x: initialPosition.x, y: initialPosition.y } }), [initialPosition]);
```
v-excelsior, line 16: Why do we use useMemo with useRef approach? 

```
position.current = onDrag({
  x: pos.x + event.movementX,
  y: pos.y + event.movementY,
});
```
        
you mutate memoized position here, isn't it?

It look absolutely wrong.

PR #765, opened 2023-09-13
MARKET-2198 [Particles] The transparency gradient doesn't work in the color setting.

serhmel:


components/bl-particles/src/helpers.js
```diff
@@ -138,6 +152,34 @@ export function useOptions(component) {
...
+  if (HSLA_REGEX.test(value)) {
+    const matches = value.match(HSLA_REGEX);
+    const color = `hsl(${ matches[1] }, ${ matches[2] }, ${ matches[3] })`;
+    const opacity = matches[4] ? parseFloat(matches[4]) : Opacity.FULL;
+
+    return { color, opacity };
+  }
+
+  return {};
+}
```
v-excelsior, line 181: If I see multiple and similar code blocks I'm asking myself if it could be rewritten. 

How about it(all names and code structure is just an example, please make it readable and meaningful):
```
const ParsersList = [
  {
    regexp          : HEX_REGEX,
    matcher         : v => v.match(HEX_REGEX),
    colorBuilder    : matches => `#${ matches[1] }`,
    opacityExtractor: matches => matches[2] ? (parseInt(matches[2], 16) / 255).toFixed(2) : Opacity.FULL,
  }
]


function parseColor(value) {
  const parser = ParsersList.find(Parser => Parser.regexp.test(value))

  if (!parser) {
    return {}
  }

  const matches = parser.matcher(value)
  const color = parser.colorBuilder(matches)
  const opacity = parser.opacityExtractor(matches)

  return { matches, color, opacity }
}
```

v-excelsior: @serhmel Testing mention in slack

PR #798, opened 2023-10-03
MARKET-2245 [Clock] The stopwatch doesn't care for the time variant.

StasivPavlo:


components/bl-clock/README.md
```diff
@@ -13,6 +13,7 @@ This is a component for Backendless [UI-Builder](https://backendless.com/develop
...
 | Time Variant: `timeVariant`                  | Select [HH:MM:SS: `hhmmss`, HH:MM: `hhmm`, HH: `hh`]            | HH:MM:SS: `hhmmss` |                          | NO           | YES        | Allows to determine variant of displaying time                                                                    |
 | Stopwatch Scale: `stopwatchScale`            | Number                                                          | 3                  |                          | NO           | YES        | Allows to determine the scale of  stopwatch seconds                                                               |
+| Stopwatch Variant: `stopwatchVariant`        | Select [SS: `ss`, MM:SS: `mmss`, HH:MM:SS: `hhmmss`]            | SS: `ss`           |                          | NO           | YES        | Allows to determine variant of displaying stopwatch                                                               |
```
v-excelsior, line 16: I'm not sure if `SS` is something meaningful for nocoders. At least first google answer for `SS` that not what I'm interested for.

components/bl-clock/src/stopwatch.js
```diff
@@ -1,20 +1,51 @@
 import { useEffect, useRef, useState } from 'react';
 
+const TimeVariant = {
+  HHMMSS: 'hhmmss',
+  MMSS: 'mmss',
+  SS: 'ss',
+};
+
+const stopwatchMap = {
```
v-excelsior, line 9: Try to find meaningful variable name

There should be an entity with which we work:
time/string/timestamp/some kind of object

and action that we do on thin data:
resolve - Resolver, format - Formatter, and so on. Composer, Converter, Transformer...

components/bl-clock/src/stopwatch.js
```diff
@@ -1,20 +1,51 @@
...
+  },
+};
+
 export function Stopwatch({ component }) {
-  const [remainingSecond, setRemainingSecond] = useState(0);
+  const { stopwatchVariant } = component;
+
+  const [second, setSeconds] = useState(0);
+  const [minutes, setMinutes] = useState(0);
+  const [hours, setHours] = useState(0);
```
v-excelsior, line 30: useObjectState?

components/bl-clock/README.md
v-excelsior, line 16: Also it looks not like variant, but `time format` or something like that.

StasivPavlo: I think it's the best name for time format. If googling "time format" everywhere is marked like this. And I think it's clear that if you're just Googling `SS`, that's probably not what you need

v-excelsior (changes requested):

components/bl-clock/src/timer.js
```diff
@@ -4,7 +4,7 @@ import { getTimer } from './helpers';
 import { Time } from './subcomponents';
 
 export function Timer({ component, eventHandlers }) {
-  const { timerDate, timeVariant, animationDuration } = component;
+  const { timerDate, animationDuration } = component;
   const { onTimerEnd } = eventHandlers;
 
   const [time, setTime] = useState(getTimer(new Date(timerDate)));
```
v-excelsior, line 10: It will be called every render

components/bl-clock/src/stopwatch.js
```diff
@@ -33,10 +62,27 @@ export function Stopwatch({ component }) {
   }, []);
 
   return (
-    <div className="stopwatch"> { remainingSecond } </div>
+    <>
+      {hasHours(stopwatchTimeFormat) && (<span className="stopwatch">{`${pad(time.hours)}:`}</span>)}
+      {hasMinutes(stopwatchTimeFormat) && (<span className="stopwatch">{`${pad(time.minutes)}:`}</span>)}
```
v-excelsior, line 67: duplicated calculations

create use memo that only once will parse time format and creates map with tokens like

```
FormatTokens = {
 HH: true,
 MM: false
}
```

v-excelsior (changes requested):

components/bl-clock/README.md
```diff
@@ -11,8 +11,9 @@ This is a component for Backendless [UI-Builder](https://backendless.com/develop
...
+| Time Format: `timeFormat`                    | Select [HH:MM:SS: `hhmmss`, HH:MM: `hhmm`, HH: `hh`]            | HH:MM:SS: `hhmmss` |                          | NO           | YES        | Allows to determine the time format                                                                               |
+| Stopwatch Scale: `stopwatchScale`            | Number                                                          | 3                  |                          | NO           | YES        | Allows to determine the scale of stopwatch seconds                                                               |
+| Stopwatch Time Format: `stopwatchTimeFormat` | Select [SS: `ss`, MM:SS: `mmss`, HH:MM:SS: `hhmmss`]            | SS: `ss`           |                          | NO           | YES        | Allows to determine the time format of displaying stopwatch                                                       |
```
v-excelsior, line 16: I'm not sure. But for this moment it looks like we join two component that works in different ways but has one common feature - time.

At least if I want to use Clock - `Stopwatch Time Format` settings confused me

@snaqqqystas @StasivPavlo what do you think if we split it for two independent components?

v-excelsior: Which PRs should we merge?

Also, I do not clearly understand why we should merge them, as they will be overwritten by new code.

snaqqqystas: i think, need to split in first on two separated components, to reduce misunderstandings

StasivPavlo: @v-excelsior Because, for example, this PR is more about the Stopwatch than the Clock component. For me, splitting the components with their existing changes will be easier than doing them again later.

Now there are no more PRs so I'm splitting the components.

v-excelsior: @StasivPavlo you should take into account that this repository is public. The entire team and customers could see this code, and we should avoid showing not-ready code.

As for the point "easier to split existing changes" - they all still exist in your local branches.

PR #806, opened 2023-10-09
(Support)[Autocomplete] Implement setting selected option

Saharij:


components/bl-autocomplete/src/helpers/value-validation.js
```diff
@@ -0,0 +1,17 @@
+export function valueValidation(value, options, hasGroup) {
```
v-excelsior, line 1: valueValidation -> validateValue/valueValidator

same for options

And make it in one file.

Also it could be a map

Validators.options(...)
Validators.value(...)

v-excelsior (changes requested):

PR #813, opened 2023-10-11
MARKET-2247 [Timer] The "On Timer End" event works wrong.

StasivPavlo:
UI-Builder: https://develop.backendless.com/app/pavlo_stasiv_components/ui-builder/Timer/pages/landing

### DIFF
--- a/components/bl-timer/src/index.js
+++ b/components/bl-timer/src/index.js
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 
 import { getTimer } from './helpers';
 import { Time } from './subcomponents';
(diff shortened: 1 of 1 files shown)


components/bl-timer/src/index.js
```diff
@@ -19,13 +19,16 @@ export default function Timer({ component, eventHandlers }) {
...
   }, [time]);
 
+  const timer = useRef(null);
+
   useEffect(() => {
-    const timer = setInterval(() => setTime(getTimer(new Date(timerDate))), 1000);
+    timer.current = setInterval(() => setTime(getTimer(new Date(timerDate))), 1000);
 
-    return () => clearInterval(timer);
+    return () => clearInterval(timer.current);
```
v-excelsior, line 27: How does it solve the issue?

Why did it run two times?
Why did it run every second after the timer finished his work?

Is it OK that the UI disappears after the timer is ended?

StasivPavlo: **_Why did it run two times?_**
First, if I'm not mistaken, this problem existed before the Clock component was separated. I suspect that since the event call condition is in the component function, it will be checked on each renderer. At the time I was doing the patching, this problem didn't exist, the event only fired once per second.

**_Why did it run every second after the timer finished his work?_**
Because the `interval `did not stop working after the event was called, accordingly, the `time` state changed every second, accordingly, a rerender occurred, which started the event again

**_Is it OK that the UI disappears after the timer is ended?_**
It's not good, so I already have a ticket for it

v-excelsior: > Because the interval did not stop working after the event was called, accordingly, the time state changed every second, accordingly, a rerender occurred, which started the event again

The means of my questions is: did you thinks it's works correctly?

For me - no. The "On Timer End" event should be fired once, in another case it should be named the "On tick after Timer End" event - which is not very convenient in general. 

> It's not good, so I already have a ticket for it
Great.

In conclusion. Let's rework the "On Timer End" event.

PR #817, opened 2023-10-11
MARKET-2271 [Stopwatch] Change the "Tick rate" option.

StasivPavlo:


components/bl-stopwatch/src/index.js
```diff
@@ -27,34 +27,24 @@ const timeFormatter = {
...
-    }
-
-    return tickRate;
-  }, []);
-
   component.start = () => {
     if (!timerRef.current) {
       const startTime = Date.now();
-      const interval = validTickRate === 0 ? 1000 : 1;
+      const interval = Number(decimalPlaces) === 0 ? 1000 : 1;
```
v-excelsior, line 39: Still not clear for me why we run this interval every 1ms if precision is 10ms or 100ms.

Is there are any reason for it?

components/bl-stopwatch/component.json
```diff
@@ -35,13 +35,31 @@
...
+        },
+        {
+          "value": "2",
+          "label": "2"
+        },
+        {
+          "value": "3",
+          "label": "3"
+        }
+      ],
```
v-excelsior, line 60:
```suggestion
      "type": "select",
      "name": "tickRate",
      "label": "Decimal Places",
      "showInSettings": true,
      "defaultValue": "1",
      "options": [
        {
          "value": "0",
          "label": "1000"
        },
        {
          "value": "1",
          "label": "100"
        },
        {
          "value": "2",
          "label": "10"
        },
        {
          "value": "3",
          "label": "1"
        }
      ],
```

Leave readable label as "Decimal Places" and use `tickRate` in code without any calculations.

StasivPavlo: I don't understand why we need to change the label if then the user will see "Decimal Places: 1000".

v-excelsior: Make mistake. Labels will be the same. Changed only values.

StasivPavlo: But for determining the number of decimal places it will not be very suitable for us. I suggest leaving value as a label, and creating a map for interval:
`
{
  '0': 1000,
  '1': 100,
  '2': 10,
  '3': 1,
};
`

v-excelsior: > it will not be very suitable for us.

Is there any arguments? 

I don't see why it could be more convenient, but if you want to leave it:

```
const TickRateInMs = {
  '0': 1000,
  '1': 100,
  '2': 10,
  '3': 1,
};
```

PR #832, opened 2023-10-23
MARKET-2286 [Stopwatch] Numbers twitches on the started stopwatch.

StasivPavlo:


components/bl-stopwatch/styles/index.less
```diff
@@ -1,9 +1,14 @@
 @bl-customComponent-stopwatch-font-size: 16px;
 @bl-customComponent-stopwatch-text-color: @appTextColor;
+@bl-customComponent-stopwatch-font-family: "Lucida Console", monospace;
 
 .bl-customComponent-stopwatch {
```
Fialkaksu, line 5: It seems to me that there is no need to separate styles; span will inherit from the parent.

```
.bl-customComponent-stopwatch {
  display: flex;
  font-size: @bl-customComponent-stopwatch-font-size;
  color: @bl-customComponent-stopwatch-text-color;
  font-family: @bl-customComponent-stopwatch-font-family;
}
```

Fialkaksu (approved):

components/bl-stopwatch/README.md
```diff
@@ -24,14 +24,10 @@ This is a component for Backendless [UI-Builder](https://backendless.com/develop
...
-```
-
-**Color**
-
-```
 @bl-customComponent-stopwatch-text-color: @appTextColor;
+@bl-customComponent-stopwatch-font-family: "Lucida Console", monospace;
 ```
+
+> **_NOTE:_** If you notice that the stopwatch twitches (this is especially noticeable when the component is centered horizontally), you may need to specify a font from the monospace family
```
v-excelsior, line 33: That does not look like a solution. Is it possible to define the char width by ourselves?

StasivPavlo: @v-excelsior  I did not find a better option. Except to wrap each number in an element and assign it a width. I got the idea [here](https://stackoverflow.com/a/10149379).

v-excelsior: @StasivPavlo Since we know how many chars we have, wrapping each element into a separate span looks not so bad to me. Do you see any cons here?

StasivPavlo: No, I will try to do it

PR #836, opened 2023-10-24
MARKET-2301 [Support - Tabs] Improve work with tabs.

StasivPavlo:


StasivPavlo:
I think changing the usage guide's last point will be better. It is used there a `getCurrentId` action to change the visibility that doesn't retrigger when the `currentTabId` changes. For example, when I determine the current tab id by logic handler, tabs aren't visible, but when I change the action to the "Tabs Content" data model in visibility logic it works. What do you think about it?

StasivPavlo (commented):
I think changing the usage guide's last point will be better. It is used there a getCurrentId action to change the visibility that doesn't retrigger when the currentTabId changes. For example, when I determine the current tab id by logic handler, tabs aren't visible, but when I change the action to the "Tabs Content" data model in visibility logic it works. What do you think about it?

v-excelsior:
> when I determine the current tab id by logic handler, tabs aren't visible

It's bug, and it should be fixed. Isn't it?

components/bl-tabs/README.md
```diff
@@ -12,18 +12,21 @@ View an example of how to install this component and how it works in your UI [he
...
 ## Events
 
-| Name                  | Triggers                        | Context Blocks               |
-|-----------------------|---------------------------------|------------------------------|
-| On Change Event       | triggered when a tab is changed | Current Tab Id: `String`     |
+| Name              | Triggers                                    | Context Blocks               |
+|-------------------|---------------------------------------------|------------------------------|
+| On Change Event   | triggered when a tab is changed             | Current Tab Id: `String`     |
+| On Mounted        | triggered after the component is mounted    |                              |
+| On Before Unmount | triggered before the component is unmounted |                              |
```
serhmel, line 29:
```suggestion
| On Mounted Event        | triggered after the component is mounted    |                              |
| On Before Unmount Event | triggered before the component is unmounted |                              |
```
?

components/bl-tabs/src/index.js
```diff
@@ -47,10 +47,18 @@ export default function TabsComponent({ component, eventHandlers, pods }) {
     }
   }, [currentTabId, tabsList]);
 
+  useEffect(() => setCurrentTabId(currentTab), [currentTab]);
+
   useEffect(() => {
     podsContent.dataModel.currentTabId = currentTabId;
```
serhmel, line 53: I do not fully understand where we use `currentTabId` from `dataModel`.
I assume that @StasivPavlo is right, and it should have been used in Visibility Logic.
But taking into account that we add `Current Tab Id`  property, I assume that we could use it instead.

v-excelsior: Now it runs on unmount too

components/bl-tabs/component.json
```diff
@@ -82,6 +82,19 @@
...
+      "name": "currentTab",
+      "label": "Current Tab Id",
+      "showInSettings": true,
+      "hasLogicHandler": true,
+      "handlerId": "currentTabLogic",
+      "handlerLabel": "Current Tab Id Logic",
+      "dataBinding": true,
+      "handlerDescription": "This handler allows you to determine the current tab.",
+      "defaultValue": "tab1",
+      "settingTooltip": "This handler allows you to determine the current tab."
```
serhmel, line 97: Don't use handler description for property tooltip.

PR #883, opened 2023-12-10
MARKET-2212 Configuration property for the Simple Modal component to control initial display state

Saharij:
https://bugs.backendless.com/browse/MARKET-2212


components/bl-simple-modal/src/utils.js
```diff
@@ -0,0 +1,10 @@
+export const modalState = {
```
v-excelsior, line 1: src/constants.js

И линтер

